Issue #7, opened 2023-02-09
SPARQL Query Errata (clarification)


afs:
https://www.w3.org/2013/sparql-errata

- [x] [clarification-query-1](https://www.w3.org/2013/sparql-errata#clarification-query-1)
- [x] [clarification-query-2](https://www.w3.org/2013/sparql-errata#clarification-query-2)
- [ ] [clarification-query-3](https://www.w3.org/2013/sparql-errata#clarification-query-3)

PR #17, opened 2023-03-03
Resolve clarification-query-3, #7

rubensworks:
This PR resolves [clarification-query-3](https://www.w3.org/2013/sparql-errata#clarification-query-3) by rephrasing the equality definition.
I also had to update some of the references, as they were pointing to outdated specs.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/17.html" title="Last updated on Mar 15, 2023, 6:10 PM UTC (31a68e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/17/7de93fb...31a68e6.html" title="Last updated on Mar 15, 2023, 6:10 PM UTC (31a68e6)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -5462,29 +5462,33 @@ class="expression">expression, ....</span>)
                                                                                    "type">RDF term</span> <span class="name">term1</span> <span class=
                                                                                                                                                 "operator">=</span> <span class="type">RDF term</span> <span class=
                                                                                                                                                                                                              "name">term2</span></pre>
-            <p>Returns TRUE if <code>term1</code> and <code>term2</code> are the same RDF term as
-              defined in [[[RDF12-CONCEPTS]]] [[RDF12-CONCEPTS]]; produces a type error if the
-              arguments are both literal but are not the same RDF term <sup><a href=
-                                                                               "#func-RDFterm-equal-foot1" class="footnote">*</a></sup>; returns FALSE otherwise.
-              <code>term1</code> and <code>term2</code> are the same if any of the following is
+            <p>Returns TRUE if <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms as defined below;
+              produces a type error if the arguments are both literals with the same datatype that is unsupported by the implementation but are not the same RDF term <sup><a href="#func-RDFterm-equal-foot1" class="footnote">*</a></sup>;
+              returns FALSE otherwise.</p>
+            <p>
+              <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms if any of the following is
               true:</p>
             <ul>
               <li>
-                <span class="name">term1</span> and <span class="name">term2</span> are equivalent
+                <span class="name">term1</span> and <span class="name">term2</span> are equal
                 <span class="IRI type">IRIs</span> as defined in <a data-cite=
-                                                                    "RDF12-CONCEPTS#section-Graph-URIref">6.4 RDF URI References</a> of
+                                                                    "RDF12-CONCEPTS#dfn-iri-equality">IRI quality</a> of
                 [[RDF12-CONCEPTS]].
               </li>
               <li>
-                <span class="name">term1</span> and <span class="name">term2</span> are equivalent
+                <span class="name">term1</span> and <span class="name">term2</span> are equal
                 <span class="literal type">literals</span> as defined in <a data-cite=
-                                                                            "RDF12-CONCEPTS#section-Literal-Equality">6.5.1 Literal Equality</a> of
+                                                                            "RDF12-CONCEPTS#dfn-term-equal">Literal term equality</a> of
                 [[RDF12-CONCEPTS]].
+              </li>
+              <li>
+                <span class="name">term1</span> and <span class="name">term2</span> are
+                <span class="literal type">literals</span> with equivalent values according to their datatypes.
               </li>
               <li>
                 <span class="name">term1</span> and <span class="name">term2</span> are the same
                 <span class="bnode type">blank node</span> as described in <a data-cite=
-                                                                              "RDF12-CONCEPTS#section-blank-nodes">6.6 Blank Nodes</a> of [[RDF12-CONCEPTS]].
+                                                                              "RDF12-CONCEPTS#section-blank-nodes">Blank Nodes</a> of [[RDF12-CONCEPTS]].
               </li>
             </ul>
             <div class="exampleGroup">
@@ -5532,7 +5536,7 @@ WHERE {
             </div>
             <p>In this query for documents that were annotated at a specific date and time (New
               Year's Day 2005, measures in timezone +00:00), the RDF terms are not the same, but have
-              equivalent values:</p>
+              equivalent values according to their datatype:</p>
             <div class="exampleGroup">
               <pre class="data">
 @prefix a:          &lt;http://www.w3.org/2000/10/annotation-ns#&gt; .
--- a/spec/index.html
+++ b/spec/index.html
@@ -5470,8 +5470,8 @@ class="expression">expression, ....</span>)
               true:</p>
             <ul>
               <li>
-                <span class="name">term1</span> and <span class="name">term2</span> are equal
-                <span class="IRI type">IRIs</span> as defined in <a data-cite=
+                <span class="name">term1</span> is an <span class="IRI type">IRI</span> and <span class="name">term2</span> is an <span class="IRI type">IRI</span>
+                such that these two <span class="IRI type">IRIs</span> are equal as per the notion of <a data-cite=
                                                                     "RDF12-CONCEPTS#dfn-iri-equality">IRI quality</a> of
                 [[RDF12-CONCEPTS]].
               </li>
--- a/spec/index.html
+++ b/spec/index.html
@@ -5462,12 +5462,14 @@ class="expression">expression, ....</span>)
                                                                                    "type">RDF term</span> <span class="name">term1</span> <span class=
                                                                                                                                                 "operator">=</span> <span class="type">RDF term</span> <span class=
                                                                                                                                                                                                              "name">term2</span></pre>
-            <p>Returns TRUE if <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms as defined below;
-              produces a type error if <code>term1</code> and <code>term2</code> are both literals that have the
+            <p>The behavior of this operator is defined as follows:</p>
+            <ul>
+                <li>Returns TRUE if <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms, as defined below.</li>
+                <li>Produces a type error if <code>term1</code> and <code>term2</code> are both literals that have the
               same <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>, this datatype IRI is <em>not</em> in the
-              set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>, and
-              the two literals are <em>not</em> the same RDF term;
-              returns FALSE otherwise.</p>
+              set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>.</li>
+                <li>Returns FALSE otherwise.</li>
+            </ul>
             <p id="func-RDFterm-equal-note1" class="note">
               An extended implementation may have support for additional datatypes. An
               implementation processing a query that tests for equivalence on non-recognized datatypes
--- a/spec/index.html
+++ b/spec/index.html
@@ -5502,10 +5502,10 @@ class="expression">expression, ....</span>)
               </li>
             </ul>
             <p id="func-RDFterm-equal-note1" class="note">
-              An extended implementation may have support for additional datatypes. An
-              implementation processing a query that tests for equivalence on non-recognized datatypes
+              An extended implementation may support additional datatypes for literals. An
+              implementation processing a query that tests for equivalence of literals with non-recognized datatypes
               (and non-identical lexical form and datatype IRI) returns an error, indicating that it
-              was unable to determine whether or not the values are equivalent. For example, an
+              is unable to determine whether or not the values of the compared literals are equivalent. For example, an
               unextended implementation will produce an error when testing either <span class=
                                                                                         "queryExcerpt"><code>"iiii"^^my:romanNumeral =
                   "iv"^^my:romanNumeral</code></span> or <span class=
--- a/spec/index.html
+++ b/spec/index.html
@@ -5466,8 +5466,18 @@ class="expression">expression, ....</span>)
               produces a type error if <code>term1</code> and <code>term2</code> are both literals that have the
               same <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>, this datatype IRI is <em>not</em> in the
               set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>, and
-              the two literals are <em>not</em> the same RDF term <sup><a href="#func-RDFterm-equal-foot1" class="footnote">*</a></sup>;
+              the two literals are <em>not</em> the same RDF term;
               returns FALSE otherwise.</p>
+            <p id="func-RDFterm-equal-note1" class="note">
+              An extended implementation may have support for additional datatypes. An
+              implementation processing a query that tests for equivalence on non-recognized datatypes
+              (and non-identical lexical form and datatype IRI) returns an error, indicating that it
+              was unable to determine whether or not the values are equivalent. For example, an
+              unextended implementation will produce an error when testing either <span class=
+                                                                                        "queryExcerpt"><code>"iiii"^^my:romanNumeral =
+                  "iv"^^my:romanNumeral</code></span> or <span class=
+                                                               "queryExcerpt"><code>"iiii"^^my:romanNumeral !=
+                  "iv"^^my:romanNumeral</code></span>.</p>
             <p>
               <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms if any of the following is
               true:</p>
@@ -5578,18 +5588,6 @@ WHERE {
                   </table>
                 </div>
               </div>
-            </div>
-            <div id="func-RDFterm-equal-foot1" class="footnote">
-              <p><sup>*</sup> Invoking RDFterm-equal on two typed literals tests for equivalent
-                values. An extended implementation may have support for additional datatypes. An
-                implementation processing a query that tests for equivalence on unsupported datatypes
-                (and non-identical lexical form and datatype IRI) returns an error, indicating that it
-                was unable to determine whether or not the values are equivalent. For example, an
-                unextended implementation will produce an error when testing either <span class=
-                                                                                          "queryExcerpt"><code>"iiii"^^my:romanNumeral =
-                    "iv"^^my:romanNumeral</code></span> or <span class=
-                                                                 "queryExcerpt"><code>"iiii"^^my:romanNumeral !=
-                    "iv"^^my:romanNumeral</code></span>.</p>
             </div>
           </section>
           <section id="func-sameTerm">
--- a/spec/index.html
+++ b/spec/index.html
@@ -5476,8 +5476,9 @@ class="expression">expression, ....</span>)
                 [[RDF12-CONCEPTS]].
               </li>
               <li>
-                <span class="name">term1</span> and <span class="name">term2</span> are equal
-                <span class="literal type">literals</span> as defined in <a data-cite=
+                <span class="name">term1</span> is a <span class="literal type">literal</span> and
+                <span class="name">term2</span> is a <span class="literal type">literal</span> such that
+                these two <span class="literal type">literals</span> are equal as per the notion of <a data-cite=
                                                                             "RDF12-CONCEPTS#dfn-term-equal">Literal term equality</a> of
                 [[RDF12-CONCEPTS]].
               </li>
--- a/spec/index.html
+++ b/spec/index.html
@@ -5470,16 +5470,6 @@ class="expression">expression, ....</span>)
               set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>.</li>
                 <li>Returns FALSE otherwise.</li>
             </ul>
-            <p id="func-RDFterm-equal-note1" class="note">
-              An extended implementation may have support for additional datatypes. An
-              implementation processing a query that tests for equivalence on non-recognized datatypes
-              (and non-identical lexical form and datatype IRI) returns an error, indicating that it
-              was unable to determine whether or not the values are equivalent. For example, an
-              unextended implementation will produce an error when testing either <span class=
-                                                                                        "queryExcerpt"><code>"iiii"^^my:romanNumeral =
-                  "iv"^^my:romanNumeral</code></span> or <span class=
-                                                               "queryExcerpt"><code>"iiii"^^my:romanNumeral !=
-                  "iv"^^my:romanNumeral</code></span>.</p>
             <p>
               <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms if any of the following is
               true:</p>
@@ -5511,6 +5501,16 @@ class="expression">expression, ....</span>)
                 <span class="bnode type">blank node</span>.
               </li>
             </ul>
+            <p id="func-RDFterm-equal-note1" class="note">
+              An extended implementation may have support for additional datatypes. An
+              implementation processing a query that tests for equivalence on non-recognized datatypes
+              (and non-identical lexical form and datatype IRI) returns an error, indicating that it
+              was unable to determine whether or not the values are equivalent. For example, an
+              unextended implementation will produce an error when testing either <span class=
+                                                                                        "queryExcerpt"><code>"iiii"^^my:romanNumeral =
+                  "iv"^^my:romanNumeral</code></span> or <span class=
+                                                               "queryExcerpt"><code>"iiii"^^my:romanNumeral !=
+                  "iv"^^my:romanNumeral</code></span>.</p>
             <div class="exampleGroup">
               <pre class="data">
 @prefix foaf:       &lt;http://xmlns.com/foaf/0.1/&gt; .
--- a/spec/index.html
+++ b/spec/index.html
@@ -5462,7 +5462,7 @@ class="expression">expression, ....</span>)
                                                                                    "type">RDF term</span> <span class="name">term1</span> <span class=
                                                                                                                                                 "operator">=</span> <span class="type">RDF term</span> <span class=
                                                                                                                                                                                                              "name">term2</span></pre>
-            <p>The behavior of this operator is defined as follows:</p>
+            <p>This function is defined as follows:</p>
             <ul>
                 <li>Returns TRUE if <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms, as defined below.</li>
                 <li>Produces a type error if <code>term1</code> and <code>term2</code> are both literals that have the
--- a/spec/index.html
+++ b/spec/index.html
@@ -5482,8 +5482,13 @@ class="expression">expression, ....</span>)
                 [[RDF12-CONCEPTS]].
               </li>
               <li>
-                <span class="name">term1</span> and <span class="name">term2</span> are
-                <span class="literal type">literals</span> with equivalent values according to their datatypes.
+                <span class="name">term1</span> is a <span class="literal type">literal</span> and
+                <span class="name">term2</span> is a <span class="literal type">literal</span>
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -5462,29 +5462,33 @@ <h5>RDFterm-equal</h5>
...
+                <span class="name">term1</span> and <span class="name">term2</span> are equal
                 <span class="literal type">literals</span> as defined in <a data-cite=
-                                                                            "RDF12-CONCEPTS#section-Literal-Equality">6.5.1 Literal Equality</a> of
+                                                                            "RDF12-CONCEPTS#dfn-term-equal">Literal term equality</a> of
                 [[RDF12-CONCEPTS]].
               </li>
+              <li>
+                <span class="name">term1</span> and <span class="name">term2</span> are
+                <span class="literal type">literals</span> with equivalent values according to their datatypes.
+              </li>
```
hartig, line 5487: I think that this new bullet point to you have added to the definition here is problematic. First of all, it is not absolutely clear what it means for literals to have "equivalent values according to their datatypes."

Here is a proposal to improve this definition:
```suggestion
              <li>
                <span class="name">term1</span> is a <span class="literal type">literal</span> and
                <span class="name">term2</span> is a <span class="literal type">literal</span>
                such that the <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>
                of these two literals is the same, this IRI is in the set of
                <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>,
                and the <a data-cite="RDF12-CONCEPTS#dfn-literal-value">literal value</a>
                associated with both literals is the same.
              </li>
```

spec/index.html
```diff
@@ -5462,29 +5462,33 @@ <h5>RDFterm-equal</h5>
...
+                <span class="name">term1</span> and <span class="name">term2</span> are equal
                 <span class="IRI type">IRIs</span> as defined in <a data-cite=
-                                                                    "RDF12-CONCEPTS#section-Graph-URIref">6.4 RDF URI References</a> of
+                                                                    "RDF12-CONCEPTS#dfn-iri-equality">IRI quality</a> of
                 [[RDF12-CONCEPTS]].
               </li>
               <li>
-                <span class="name">term1</span> and <span class="name">term2</span> are equivalent
+                <span class="name">term1</span> and <span class="name">term2</span> are equal
                 <span class="literal type">literals</span> as defined in <a data-cite=
```
hartig, line 5480:
```suggestion
                <span class="name">term1</span> is a <span class="literal type">literal</span> and
                <span class="name">term2</span> is a <span class="literal type">literal</span> such that
                these two <span class="literal type">literals</span> are equal as per the notion of <a data-cite=
```

spec/index.html
```diff
@@ -5462,29 +5462,33 @@ <h5>RDFterm-equal</h5>
...
               </li>
+              <li>
+                <span class="name">term1</span> and <span class="name">term2</span> are
+                <span class="literal type">literals</span> with equivalent values according to their datatypes.
+              </li>
               <li>
                 <span class="name">term1</span> and <span class="name">term2</span> are the same
                 <span class="bnode type">blank node</span> as described in <a data-cite=
-                                                                              "RDF12-CONCEPTS#section-blank-nodes">6.6 Blank Nodes</a> of [[RDF12-CONCEPTS]].
+                                                                              "RDF12-CONCEPTS#section-blank-nodes">Blank Nodes</a> of [[RDF12-CONCEPTS]].
```
hartig, line 5491: Currently, the section that you reference here does not talk about "same" blank nodes. However, I think it is okay to remove the reference; i.e., just end the sentence after: "[...] same blank node."

hartig (changes requested):

spec/index.html
rubensworks: Ok, I agree. Removed in the latest commit.

spec/index.html
```diff
@@ -5462,29 +5462,38 @@ <h5>RDFterm-equal</h5>
...
                                                                                                                                                                                                              "name">term2</span></pre>
-            <p>Returns TRUE if <code>term1</code> and <code>term2</code> are the same RDF term as
-              defined in [[[RDF12-CONCEPTS]]] [[RDF12-CONCEPTS]]; produces a type error if the
-              arguments are both literal but are not the same RDF term <sup><a href=
-                                                                               "#func-RDFterm-equal-foot1" class="footnote">*</a></sup>; returns FALSE otherwise.
-              <code>term1</code> and <code>term2</code> are the same if any of the following is
+            <p>Returns TRUE if <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms as defined below;
+              produces a type error if the arguments are both literals with the same datatype that is unsupported by the implementation but are not the same RDF term <sup><a href="#func-RDFterm-equal-foot1" class="footnote">*</a></sup>;
```
hartig, line 5466:
```suggestion
              produces a type error if <code>term1</code> and <code>term2</code> are both literals that have the
              same <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>, this datatype IRI is <em>not</em> in the
              set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>, and
              the two literals are <em>not</em> the same RDF term <sup><a href="#func-RDFterm-equal-foot1" class="footnote">*</a></sup>;
```

hartig (changes requested):

spec/index.html
```diff
@@ -5462,29 +5462,38 @@ <h5>RDFterm-equal</h5>
...
+                <span class="name">term2</span> is a <span class="literal type">literal</span> such that
+                these two <span class="literal type">literals</span> are equal as per the notion of <a data-cite=
+                                                                            "RDF12-CONCEPTS#dfn-term-equal">Literal term equality</a> of
                 [[RDF12-CONCEPTS]].
               </li>
+              <li>
+                <span class="name">term1</span> is a <span class="literal type">literal</span> and
+                <span class="name">term2</span> is a <span class="literal type">literal</span>
+                such that the <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>
+                of these two literals is the same, this IRI is in the set of
```
afs, line 5489:
```suggestion
                of these two literals is the same, both datatype IRIs are in the set of the
```

The datatypes can be different (example `xsd:integer` and `xsd:decimal`, or `xsd:integer` and `xsd:double`).

spec/index.html
```diff
@@ -5462,29 +5462,38 @@ <h5>RDFterm-equal</h5>
...
+                <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>,
+                and the <a data-cite="RDF12-CONCEPTS#dfn-literal-value">literal value</a>
```
afs, line 5491:
```suggestion
                and the <a data-cite="RDF12-CONCEPTS#dfn-literal-value">literal values</a>
```

spec/index.html
```diff
@@ -5462,29 +5462,38 @@ <h5>RDFterm-equal</h5>
...
+                associated with both literals is the same.
```
afs, line 5492:
```suggestion
                associated with both literals are the same value.
```

afs (changes requested):
Overall, looking good. This is a hard area to write text for that is comprehensible and correct.

spec/index.html
hartig: @afs Isn't the case of different datatypes relevant only for numeric datatypes? If so, then I think we don't need to cover this case here because equivalence testing (i.e., the `=` operator) for numerics is not defined in terms of the `RDFterm-equal` function. Instead, the [operator mapping table in Sec.17.3](https://w3c.github.io/sparql-query/spec/#OperatorMapping) defines `=` for numerics using the [XPath function `op:numeric-equal`](https://www.w3.org/TR/xpath-functions/#func-numeric-equal).

hartig:
I would propose the following additional change to improve this section further, which would also address another minor aspect of the [original bug report](https://lists.w3.org/Archives/Public/public-rdf-dawg-comments/2018Aug/0000.html) that led to the [clarification-query-3](https://www.w3.org/2013/sparql-errata#clarification-query-3) erratum. The author of the bug report is confused about "the hyphen preceding the semicolon." The issue is that there actually is no hyphen in the text but, instead, the asterisk (*) that links to the footnote is underlined because it is a hyperlink, but that underlining looks a bit like a hyphen. Now, my proposal is to get rid of the footnote and move the text of the footnote into a Note block (`<p class="note"> ... </p>`) that we place between the fourth/last bullet point and the first example in this section. Good idea?

spec/index.html
```diff
@@ -5462,29 +5462,41 @@ <h5>RDFterm-equal</h5>
...
                                                                                                                                                 "operator">=</span> <span class="type">RDF term</span> <span class=
                                                                                                                                                                                                              "name">term2</span></pre>
-            <p>Returns TRUE if <code>term1</code> and <code>term2</code> are the same RDF term as
-              defined in [[[RDF12-CONCEPTS]]] [[RDF12-CONCEPTS]]; produces a type error if the
-              arguments are both literal but are not the same RDF term <sup><a href=
-                                                                               "#func-RDFterm-equal-foot1" class="footnote">*</a></sup>; returns FALSE otherwise.
-              <code>term1</code> and <code>term2</code> are the same if any of the following is
+            <p>Returns TRUE if <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms as defined below;
```
rubensworks, line 5465: This sentence has become very long now.
I propose to split this up in an `<ul>`.
Any concerns about this?

rubensworks:
> Now, my proposal is to get rid of the footnote and move the text of the footnote into a Note block (<p class="note"> ... </p>) that we place between the fourth/last bullet point and the first example in this section. Good idea?

👍 

Feel free to delegate this to me if you want me to incorporate that into this PR.

hartig:
@rubensworks 

> Feel free to delegate this to me if you want me to incorporate that into this PR.

Yes please do. Thanks in advance!

spec/index.html
afs: This text has to be general, including for additional datatypes. 

In extended types, xsd:duration has derived datatypes. Somewhere, there could be "astromonical time" mapping to the timeline, or datatypes definitions mapping to the RGB color space.

This defines the extensions of `=`, including knowing when value spaces are distinct and `=` is false.

hartig: I was assuming that such special cases (i.e., cases with different datatypes but equivalent literal values) can be captured formally as SPARQL language extensions that are defined by adding further rows to the operator mapping table in Sec.17.3 rather than relying on the definition of the `RDFterm-equal` function. At least, that's my interpretation of the first paragraph in [Sec. 17.3.1 Operator Extensibility](https://w3c.github.io/sparql-query/spec/#operatorExtensibility).

However, if such cases were/are meant to be captured through `RDFterm-equal` as you describe it, then I think the sentence that we are discussing here needs to be improved to make this aspect more explicit. Currently, the sentence (with your suggested change already applied) says:

> term1 is a literal and term2 is a literal such that the [datatype IRI](https://w3c.github.io/rdf-concepts/spec/#dfn-datatype-iri) of these two literals is the same, both datatype IRIs are in the set of the [recognized datatype IRIs](https://w3c.github.io/rdf-concepts/spec/#dfn-recognized-datatype-iri), and the [literal values](https://w3c.github.io/rdf-concepts/spec/#dfn-literal-value) associated with both literals is the same.

To capture your interpretation we need two changes in this sentence:
1. The first condition ("such that the [datatype IRI](https://w3c.github.io/rdf-concepts/spec/#dfn-datatype-iri) of these two literals is the same") has to be removed. (Do you agree?)
2. The last condition ("the [literal values](https://w3c.github.io/rdf-concepts/spec/#dfn-literal-value) associated with both literals is the same") needs to be changed because---if I understand your interpretation right (let me know if not)---it is actually okay if the two literal values are not the same value, as long as they are considered as equivalent values.

Is this what you are suggesting?

rubensworks:
@hartig 

> > Feel free to delegate this to me if you want me to incorporate that into this PR.

> Yes please do. Thanks in advance!

Done in ce5d78351c685a4d8b9dc47fad5d600cc14974ac

spec/index.html
rubensworks: Done in 15e5f4ebe3941b1a486fc5f6c204d9f6a38a49fb

hartig:
@rubensworks 
> Done in https://github.com/w3c/sparql-query/commit/ce5d78351c685a4d8b9dc47fad5d600cc14974ac

My proposal was to place the note  between the fourth/last bullet point (of the definition of equal RDF terms) and the first example. Can you please move it there.

spec/index.html
```diff
@@ -5462,29 +5462,53 @@ <h5>RDFterm-equal</h5>
...
                                                                                                                                                 "operator">=</span> <span class="type">RDF term</span> <span class=
                                                                                                                                                                                                              "name">term2</span></pre>
-            <p>Returns TRUE if <code>term1</code> and <code>term2</code> are the same RDF term as
-              defined in [[[RDF12-CONCEPTS]]] [[RDF12-CONCEPTS]]; produces a type error if the
-              arguments are both literal but are not the same RDF term <sup><a href=
-                                                                               "#func-RDFterm-equal-foot1" class="footnote">*</a></sup>; returns FALSE otherwise.
-              <code>term1</code> and <code>term2</code> are the same if any of the following is
+            <p>The behavior of this operator is defined as follows:</p>
```
hartig, line 5465:
```suggestion
            <p>This function is defined as follows:</p>
```
Notice that we are not actually defining the `=` operator here but, instead, the `RDFterm-equal` function. That can be a bit confusing because the signature in the beginning of this section seems to indicate that this section is about the `=` operator. There is an open issue, #18, with a proposal to change that part of the section in order to avoid this confusion.

rubensworks:
@hartig 

> My proposal was to place the note between the fourth/last bullet point (of the definition of equal RDF terms) and the first example. Can you please move it there.

Done.

spec/index.html
```diff
@@ -5462,29 +5462,53 @@ <h5>RDFterm-equal</h5>
...
+            <ul>
+                <li>Returns TRUE if <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms, as defined below.</li>
+                <li>Produces a type error if <code>term1</code> and <code>term2</code> are both literals that have the
+              same <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>, this datatype IRI is <em>not</em> in the
+              set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>.</li>
```
hartig, line 5470: Notice that these two points are not mutually exclusive and their order matters. I would try to avoid that and make sure that every case fits only one of these points (which would then make the order of these points irrelevant).

The issue is with cases in which the two terms are literals that have i) the same datatype IRI, ii) this IRI is not one of the recognized datatype IRIs, and iii) the two literals have the same lexical form (i.e., they are actually the same literal). This case is covered by the first point, but also by the second one. By the first point, the function returns TRUE in this case, and I think that this is the expected behavior.

Now, to make sure that this case does not also fit the second point (and, thus, to make the order of the points irrelevant) I propose to rephrase the second point as follows.
```suggestion
                <li>Returns TRUE if <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms, as defined below.</li>
                <li>Produces a type error if <code>term1</code> and <code>term2</code> are both literals such that these two literals have the
              same <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>, this datatype IRI is <em>not</em> in the
              set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>, and the
              two literals have the same <a data-cite="RDF12-CONCEPTS#dfn-lexical-form">lexical form</a>.</li>
```

spec/index.html
```diff
@@ -5462,31 +5462,55 @@ <h5>RDFterm-equal</h5>
...
-                <span class="bnode type">blank node</span> as described in <a data-cite=
-                                                                              "RDF12-CONCEPTS#section-blank-nodes">6.6 Blank Nodes</a> of [[RDF12-CONCEPTS]].
+                <span class="bnode type">blank node</span>.
               </li>
             </ul>
+            <p id="func-RDFterm-equal-note1" class="note">
+              An extended implementation may have support for additional datatypes. An
+              implementation processing a query that tests for equivalence on non-recognized datatypes
+              (and non-identical lexical form and datatype IRI) returns an error, indicating that it
+              was unable to determine whether or not the values are equivalent. For example, an
```
hartig, line 5508:
```suggestion
              An extended implementation may support additional datatypes for literals. An
              implementation processing a query that tests for equivalence of literals with non-recognized datatypes
              (and non-identical lexical form and datatype IRI) returns an error, indicating that it
              is unable to determine whether or not the values of the compared literals are equivalent. For example, an
```

spec/index.html
rubensworks: I agree, making the different cases mutually exclusive would make things clearer.

However, instead of

> two literals have the same <a data-cite="RDF12-CONCEPTS#dfn-lexical-form">lexical form</a>.

in the second case, I suspect you mean

> two literals a different <a data-cite="RDF12-CONCEPTS#dfn-lexical-form">lexical form</a>.

?

hartig: Yes, that's what I meant ;-)

Here is the updated suggestion, ready for you to commit.
```suggestion
                <li>Returns TRUE if <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms, as defined below.</li>
                <li>Produces a type error if <code>term1</code> and <code>term2</code> are both literals such that these two literals have the
              same <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>, this datatype IRI is <em>not</em> in the
              set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>, and the
              <a data-cite="RDF12-CONCEPTS#dfn-lexical-form">lexical forms</a> of the two literals are different from one another.</li>
```

spec/index.html
```diff
@@ -5465,9 +5465,10 @@ <h5>RDFterm-equal</h5>
             <p>This function is defined as follows:</p>
             <ul>
                 <li>Returns TRUE if <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms, as defined below.</li>
-                <li>Produces a type error if <code>term1</code> and <code>term2</code> are both literals that have the
+                <li>Produces a type error if <code>term1</code> and <code>term2</code> are both literals such that these two literals have the
               same <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>, this datatype IRI is <em>not</em> in the
-              set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>.</li>
+              set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>, and the
+              <a data-cite="RDF12-CONCEPTS#dfn-lexical-form">lexical forms</a> of the two literals are different from one another.</li>
```
```suggestion
                <li>Produces a type error if <code>term1</code> and <code>term2</code> are both literals having the
              same <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>; this datatype IRI is <em>not</em> in the
              set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>; and the
              <a data-cite="RDF12-CONCEPTS#dfn-lexical-form">lexical forms</a> of the two literals are different from one another.</li>
```

spec/index.html
hartig: @afs please take a look at my previous comment above once you have time.

spec/index.html
rubensworks: @hartig Since you suggested this text before, do you agree with this change? (feel free to commit directly if so)

spec/index.html
```diff
@@ -5462,31 +5462,56 @@ <h5>RDFterm-equal</h5>
...
-              arguments are both literal but are not the same RDF term <sup><a href=
-                                                                               "#func-RDFterm-equal-foot1" class="footnote">*</a></sup>; returns FALSE otherwise.
-              <code>term1</code> and <code>term2</code> are the same if any of the following is
+            <p>This function is defined as follows:</p>
+            <ul>
+                <li>Returns TRUE if <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms, as defined below.</li>
+                <li>Produces a type error if <code>term1</code> and <code>term2</code> are both literals such that these two literals have the
+              same <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>, this datatype IRI is <em>not</em> in the
+              set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>, and the
+              <a data-cite="RDF12-CONCEPTS#dfn-lexical-form">lexical forms</a> of the two literals are different from one another.</li>
```
TallTed, line 5471:
```suggestion
                <li>Produces a type error if <code>term1</code> and <code>term2</code> are both literals having the
              same <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>; this datatype IRI is <em>not</em> in the
              set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>; and the
              <a data-cite="RDF12-CONCEPTS#dfn-lexical-form">lexical forms</a> of the two literals are different from one another.</li>
```

spec/index.html
TallTed: @rubensworks @hartig 

GitHub seems to be confused about this PR. The conversation [seen here, in the _Conversation_ tab](https://github.com/w3c/sparql-query/pull/17) does not track with the conversation [seen on the _Files changed_ tab](https://github.com/w3c/sparql-query/pull/17/files).

I have resuggested the same change, against the current state of the file. Maybe [this](https://github.com/w3c/sparql-query/pull/17#pullrequestreview-1339564323) can be merged (and [this](https://github.com/w3c/sparql-query/pull/17#discussion_r1132788329) can be marked resolved)?

spec/index.html
hartig: Thanks @TallTed for reproducing your suggested change!

spec/index.html
hartig: I am marking [this change suggestion](https://github.com/w3c/sparql-query/pull/17#discussion_r1132788329) as resolved by merging [the other one](https://github.com/w3c/sparql-query/pull/17#pullrequestreview-1339564323).

spec/index.html
afs: It might be better to rename the function. `RDFterm-equal` puts too much on the "RDFterm" and so reads like 'sameTerm' which it isn't.

It should be "operator(RDFterm,RDFTerm)-equality-by-value" which is ungainly :-)

"equals" is used for same-value (equal in datatype range), to contrast with `sameTerm` (equal in datatype domain).

Suggestions for alternative names:
* `RDF-value-equals`
* `sameValue` (c.f. `sameTerm`)
* `value-equals`

Yes, the first condition has to be removed.

I don't follow your point 2. How is "same value" different from "equivalent values"? Two values are the same if and only if they are identical - same elements of the set that is the common range of the two datatype mappings.

The original "associated with both literals" was meant to mean the values of each datatype mapping.

`sameValue` for literals is whether `equality(  datatype1(lexical1) , datatype2(lexcial2) )` where `equality` is a partial function of identity in the common range of the datatype mappings - the intersection of the two value spaces of datatype1 and datatype2.

Value spaces may overlap - intersection.

If the values for the literals are not both in the common range, then return "false" without knowing anything more. A common case is where the values spaces are disjoint.

`sameValue` is different to other extensions in that it gets the heart of what a literal with datatype is in RDF. Extensions need to follow that.


Can we get some adequate text now, get the working draft updated, and then loop round to discussing changes to the updated text, including a longer discussion at the end of 17.4.1.7 (we have multiple audiences - users and implementers ... as well as spec-ites)?

spec/index.html
```diff
@@ -5462,31 +5462,56 @@ <h5>RDFterm-equal</h5>
...
+                <span class="name">term2</span> is a <span class="literal type">literal</span> such that
+                these two <span class="literal type">literals</span> are equal as per the notion of <a data-cite=
+                                                                            "RDF12-CONCEPTS#dfn-term-equal">Literal term equality</a> of
                 [[RDF12-CONCEPTS]].
               </li>
+              <li>
+                <span class="name">term1</span> is a <span class="literal type">literal</span> and
+                <span class="name">term2</span> is a <span class="literal type">literal</span>
+                such that the <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>
+                of these two literals is the same, this IRI is in the set of
```
hartig, line 5495: Based on the discussion [here](https://github.com/w3c/sparql-query/pull/17/files#r1136897998), I am changing the text as follows.
```suggestion
                such that the <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>
                of each of these two literals is in the set of
```

spec/index.html
```diff
@@ -5462,31 +5462,56 @@ <h5>RDFterm-equal</h5>
...
                 [[RDF12-CONCEPTS]].
               </li>
+              <li>
+                <span class="name">term1</span> is a <span class="literal type">literal</span> and
+                <span class="name">term2</span> is a <span class="literal type">literal</span>
+                such that the <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>
+                of each of these two literals is in the set of
+                <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>,
+                and the <a data-cite="RDF12-CONCEPTS#dfn-literal-value">literal values</a>
+                associated with both literals are the same value.
```
hartig, line 5498:
```suggestion
                <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>
                and both literals have the <a data-cite="RDF12-CONCEPTS#dfn-literal-value">literal value</a>.
```

spec/index.html
hartig: @afs Regarding your proposal to rename `RDFterm-equal`, I agree. I have created a separate issue for it: #25

spec/index.html
```diff
@@ -5492,10 +5492,9 @@ <h5>RDFterm-equal</h5>
                 <span class="name">term1</span> is a <span class="literal type">literal</span> and
                 <span class="name">term2</span> is a <span class="literal type">literal</span>
                 such that the <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>
-                of these two literals is the same, this IRI is in the set of
-                <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>,
-                and the <a data-cite="RDF12-CONCEPTS#dfn-literal-value">literal values</a>
-                associated with both literals are the same value.
+                of each of these two literals is in the set of
+                <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>
+                and both literals have the <a data-cite="RDF12-CONCEPTS#dfn-literal-value">literal value</a>.
```
TallTed, line 5497:
```suggestion
                and both literals have the same <a data-cite="RDF12-CONCEPTS#dfn-literal-value">literal value</a>.
```

spec/index.html
```diff
@@ -5462,31 +5462,55 @@ <h5>RDFterm-equal</h5>
...
+                these two <span class="literal type">literals</span> are equal as per the notion of <a data-cite=
+                                                                            "RDF12-CONCEPTS#dfn-term-equal">Literal term equality</a> of
                 [[RDF12-CONCEPTS]].
               </li>
+              <li>
+                <span class="name">term1</span> is a <span class="literal type">literal</span> and
+                <span class="name">term2</span> is a <span class="literal type">literal</span>
+                such that the <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>
+                of each of these two literals is in the set of
+                <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>
```
TallTed, line 5496: Is this intentionally dropping `the datatype IRI of the two literals is the same`? It doesn't seem like they should differ between the two literals....
```suggestion
                of each of these two literals is the same; this 
                <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a> is in the set of
                <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>;
```

spec/index.html
hartig: Oops! Thanks!!

spec/index.html
TallTed: Ah, got it. Anyone who has such a button (I don't) can mark this Resolved.

hartig: Marking this as resolved.

spec/index.html
afs: > If this is okay with you

Yes - that's good for merge. 

We (and I offer to do it) can revisit when we have a new baseline text to work from alongside #25.

PR #37, opened 2023-03-25
errata-query-23

afs:
https://www.w3.org/2013/sparql-errata#errata-query-23

https://lists.w3.org/Archives/Public/public-rdf-dawg-comments/2023Jan/0000.html

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -9722,7 +9722,7 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
           <section id="defn_aggCount">
             <h5>Count</h5>
             <p>Count is a SPARQL set function which counts the number of times a given expression
-              has a bound, and non-error value within the aggregate group.</p>
+              has a bound, and non-error, value within the aggregate group.</p>
             <div class="defn">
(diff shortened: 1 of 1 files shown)


hartig (approved):

domel (approved):

spec/index.html
```diff
@@ -9722,7 +9722,7 @@ <h5>Set Functions</h5>
           <section id="defn_aggCount">
             <h5>Count</h5>
             <p>Count is a SPARQL set function which counts the number of times a given expression
-              has a bound, and non-error value within the aggregate group.</p>
+              has a bound, and non-error, value within the aggregate group.</p>
```
rubensworks, line 9725: The placement of this new comma reads a bit confusing now.

Alternatively, we could choose to not add this comma, and also remove the comma after `bound`.

afs: How about simply "has a value within the aggregate group." 

It says a few lines down in the defn "remove error elements from N".

rubensworks (approved):

PR #43, opened 2023-03-27
fix action version

pchampin:


afs:
Could you delete the branch please?

Branches can be restored.

Because we are using squash-and-merge, a branch will always show as having commits not present on main. If left open, it makes it hard to see what's completed and what's not when looking at the branch listing.

PR #44, opened 2023-03-29
Inline local.css for 'preview'

afs:
This is an attempt to get preview working for PR's. 

The problem seems to be that local.css is not being placed into the PR preview folder.

With this PR, the preview should look better.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/44.html" title="Last updated on Mar 29, 2023, 10:24 AM UTC (669b878)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/44/d4bc873...669b878.html" title="Last updated on Mar 29, 2023, 10:24 AM UTC (669b878)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -50,7 +50,265 @@
     </script>
 
     <style>
-      @import url("local.css");
+      /* @import url("local.css"); */
+      /* Inlined to make preview work */
+
+/* CSS For SPARQL Query */
+
+/* In-progress working draft artifacts - to be removed eventually */
+  .issue	{ background-color: #fdd;
+                  font-size: 88% ; }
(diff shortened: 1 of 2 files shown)


afs:
Looks better for me.

If this is the case, we can do the same for the other SPARQL repos as a "quick fix" until proper reformatting can be done.

It does not fix the general inclusion problem: there is still:

```
Cannot GET /uploads/Aifol7/spec/common/sparql-related.html
```

hartig (approved):
Good idea ;-)  Looks better for me too.

afs:
Thanks for confirming it looks better. I can roll out a similar quick fix for other docs.

PR #50, opened 2023-04-05
now issues are also added to the project

pchampin:


afs (approved):
Feel free to consider this as an approval for other sparql-* repos as well.

PR #71, opened 2023-05-07
improve display on mobile phones

domel:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/71.html" title="Last updated on May 16, 2023, 9:51 PM UTC (5eda33a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/71/c1bcb0b...5eda33a.html" title="Last updated on May 16, 2023, 9:51 PM UTC (5eda33a)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -333,10 +333,9 @@ span.cancast:hover { background-color: #ffa;
       em.rfc2119 { text-transform: lowercase;
                    font-variant: small-caps;
                    font-style: normal; }
-      table { word-break: break-all; max-width: 80vw; }
-      table th:nth-child(1), table td:nth-child(1) { padding: .2em; min-width: 4vw; }
-      table th:nth-child(2), table td:nth-child(2) { padding: .2em; min-width: 10vw; }
-      table th:nth-child(3), table td:nth-child(3) { padding: .2em; min-width: 4vw; }
+      @media (max-width: 767px) {
+        table { word-break: break-all; max-width: 80vw; }
+      }
     </style>
   </head>
   <body>
--- a/spec/index.html
+++ b/spec/index.html
@@ -335,6 +335,8 @@ span.cancast:hover { background-color: #ffa;
                    font-style: normal; }
       @media (max-width: 767px) {
         table { word-break: break-all; max-width: 80vw; }
+        .grammarTable table td:nth-child(1) { min-width: 12.5vw; }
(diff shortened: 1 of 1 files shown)


rubensworks:
See my comment in sparql-protocol: https://github.com/w3c/sparql-protocol/pull/15#issuecomment-1537829179

hartig:
@domel Do you have a concrete example where this change makes a difference and what this difference looks like?

spec/index.html
```diff
@@ -333,10 +333,9 @@
       em.rfc2119 { text-transform: lowercase;
                    font-variant: small-caps;
                    font-style: normal; }
-      table { word-break: break-all; max-width: 80vw; }
-      table th:nth-child(1), table td:nth-child(1) { padding: .2em; min-width: 4vw; }
-      table th:nth-child(2), table td:nth-child(2) { padding: .2em; min-width: 10vw; }
-      table th:nth-child(3), table td:nth-child(3) { padding: .2em; min-width: 4vw; }
+      @media (max-width: 767px) {
+        table { word-break: break-all; max-width: 80vw; }
```
afs, line 337: Would `break-word` be better? It copes with the intermediate cases better for me - e.g. PC-narrow window, and mobile/landscape.

afs: So normal+overflow-wrap is the same effect? Do those give word wrapping? 

`break-all` breaks within words when a word break is possible.

TallTed: > [@domel] `break-word` is deprecated

Do you have a link to this deprecation and the subsequent recommendations?

I think we should NOT use anything that breaks mid-word unless it's using some sort of hyphenation dictionary, and even then, I think we should be reviewing the small-screen results carefully (and advising readers, somewhere, somehow, to use bigger rendering environments to best consume these documents).

domel: > > [@domel] `break-word` is deprecated
> 
> Do you have a link to this deprecation and the subsequent recommendations?

[CSS Text Module Level 3](https://www.w3.org/TR/css-text-3/#word-break):

> For compatibility with legacy content, the word-break property also supports a deprecated break-word keyword. When specified, this has the same effect as word-break: normal and overflow-wrap: anywhere, regardless of the actual value of the overflow-wrap property.

afs (approved):

PR #79, opened 2023-05-19
Move RDFterm-equal and sameTerm to term functions section

### LINKED ISSUE #67: Functional Forms section also has proper functions in it
[Functional Forms](https://w3c.github.io/sparql-query/spec/index.html#func-forms)

Currently (May 2023):
17.4.1.1 BOUND
17.4.1.2 IF
17.4.1.3 COALESCE
17.4.1.4 NOT EXISTS and EXISTS
17.4.1.5 logical-or
17.4.1.6 logical-and
17.4.1.7 RDFterm-equal
17.4.1.8 sameTerm
17.4.1.9 IN
17.4.1.10 NOT IN

`RDFterm-equal`, `sameTerm`, are normal functions that evaluate all their arguments then pass the eval results as args to the function. They can be be moved to the next section "Functions on RDF Terms".

afs:
This resolves #67.

This PR moves `RDFterm-equal` and `sameTerm` from the [Functional Forms](https://github.com/w3c/sparql-query/compare/term-func?expand=1#func-forms) section to [Functions on RDF Terms](https://github.com/w3c/sparql-query/compare/term-func?expand=1#func-rdfTerms) section. It also puts `sameTerm` first, swapping the order of `RDFterm-equal` and `sameTerm`.

The PR makes no other changes other than moving these blocks of text.

The git changes shows as if `IN` and `NOT IN` have been moved - they haven't. It's seems git calculated the changeset in a different way.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -5554,6 +5554,246 @@ class="expression">expression, ....</span>)
             <p>Note: see section 17.2, <a href="#evaluation">Filter Evaluation</a>, for the
               <code>&amp;&</code> operator's treatment of errors.</p>
           </section>
+          <section id="func-in">
+            <h5>IN</h5>
+            <pre class="prototype nohighlight">
+<span class="return">boolean</span>  rdfTerm <span class="operator">IN</span> (<span class="expression">expression</span>, <span class="expression">...</span>)
+            </pre>
+            <p>
+              The <code>IN</code> operator tests whether the RDF term on the
+              left-hand side is found in the list of values of the expressions
+              on the right-hand side. The test is done with the "=" operator,
+              which tests for the same value, as determined by the
+              <a href="#OperatorMapping">operator mapping</a>.
+            </p>
+            <p>
+              A list of zero terms on the right-hand side is legal and evaluates
+              to <code>false</code>.
+            </p>
+            <p>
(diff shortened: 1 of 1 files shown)


rubensworks (approved):

hartig:
I am fine with moving these subsections from the "functional forms" section to the "functions" section as done in this PR. One thing that I am not entirely happy with, however, is that the subsection about `RDFterm-equal` is now in between the subsections of all the other functions as if `RDFterm-equal` was an ordinary function like all the others, which it is not. In contrast to the other functions, `RDFterm-equal` is not a function that can actually be used in expressions; instead, its sole purpose is for defining the semantics of the `=` operator. Therefore, I would prefer that this special role of `RDFterm-equal` was also reflected by the placement of the corresponding subsection. For instance, it could be the very first subsection (i.e., 17.4.2.1) or the very last one (i.e., 17.4.2.15).

(I notice that the subsections for `logical-or` and `logical-and` have the same issue.)

afs:
Having `RDFterm-equal` as 17.4.2.1 would be OK - I'll move it.

To me, the important thing is that it is a proper function with all the evaluation and optimization consequences that follow.

hartig (approved):

PR #97, opened 2023-06-10
Removes the historical notes on rdf:langString datatype

Tpt:
rdf:langString is normative since RDF 1.1 and there is already a note just above to highlight rdf:langString usage


afs (approved):

domel (approved):

hartig (approved):

afs:
I think we can ignore the Echidna error.

Issue #95, opened 2023-06-09
Another issue related to 'errata-query-11' (sequence versus multiset)


hartig:
When looking at [errata-query-11](https://www.w3.org/2013/sparql-errata#errata-query-11), another question came to my mind. The suggest fix for the errata also proposes to

> replace "multiset of partial functions" by "set of partial functions"

in the [definition of the Aggregation operator](https://www.w3.org/TR/sparql11-query/#defn_algAggregation). In particular, this proposal is about the first line of the definition, which ends as follows.

"[...] and let { key1→Ω1, ..., keym→Ωm } be a multiset of partial functions from keys to solution sequences as produced by the grouping step."

I certainly agree with the proposal to replace "multiset" by "set". However, I wonder why the definition talks about "solution sequences"?

While I see that the corresponding definition of "the grouping step"---that is, the [definition of the Group operator](https://www.w3.org/TR/sparql11-query/#defn_algGroup)---also says that it "produc[es] a set of partial functions from keys to solution sequences," the partial functions produced by the actual formula of this definition (as copied below) are from keys to _multisets (or sets?)_ of solution mappings, not to sequences of solution mappings.
```
Group(exprlist, Ω) = { ListEval(exprlist, μ) → { μ' | μ' in Ω, ListEval(exprlist, μ) = ListEval(exprlist, μ') } | μ in Ω }
```
(Notice that `Ω` is a multiset of solution mappings because that is passed to the [Group operator](https://www.w3.org/TR/sparql11-query/#defn_algGroup) by the [Evaluation of Group](https://www.w3.org/TR/sparql11-query/#defn_evalGroup), where it is obtained from `eval(D(G), P)` with `P` being "the algebra translation of the GroupGraphPattern of the query level" according to the translation algorithm for aggregates in [Section 18.2.4.1 Grouping and Aggregation](https://www.w3.org/TR/sparql11-query/#sparqlGroupAggregate)).

Consequently, in addition to replacing "multiset" by "set", I propose to also replace "solution sequences" by "sets of solution mappings", and this latter replacement should be done not only in the [definition of the Aggregation operator](https://www.w3.org/TR/sparql11-query/#defn_algAggregation) but also in the [definition of the Group operator](https://www.w3.org/TR/sparql11-query/#defn_algGroup).

kasei:
[...]

Issue #94, opened 2023-06-09
Issue with 'errata-query-11' (no need to change 'Evaluation of Aggregation')


hartig:
I am looking at [errata-query-11](https://www.w3.org/2013/sparql-errata#errata-query-11). The suggested fix is that ...
> The definition of Aggregation() should replace P by Group(exprlist, P) giving:
> ```
> eval(D(G), Aggregation(exprlist, func, scalarvals, Group(exprlist, P))) =
>        Aggregation(exprlist, func, scalarvals, eval(D(G), Group(exprlist, P)))
> ```

[The current version of the definition](https://www.w3.org/TR/sparql11-query/#defn_evalAggregation) looks as follows.

```
eval(D(G), Aggregation(exprlist, func, scalarvals, P)) =
       Aggregation(exprlist, func, scalarvals, eval(D(G), P))
```

While I can see that using the symbol `P` in this definition can cause confusion, I don't think there is anything wrong with the current definition per se. Here is why:

Given how any expression of the form `Aggregation(exprlist, func, scalarvals, P)` is produced by the translation algorithm in [Section 18.2.4.1 Grouping and Aggregation](https://www.w3.org/TR/sparql11-query/#sparqlGroupAggregate), it is guaranteed that `P` is an expression of the form `Group(exprlist, P')` where `P'` is "the algebra translation of the GroupGraphPattern of the query level" (i.e., also an expression). The part that may cause the confusion is that the symbol `P` here is named `G` in the translation algorithm in [Section 18.2.4.1 Grouping and Aggregation](https://www.w3.org/TR/sparql11-query/#sparqlGroupAggregate) (and our `P'` is named `P` in the algorithm).

[...]

PR #98, opened 2023-06-14
Turn the input to Group(..) and to Aggregation(..) into solution sequences

hartig:
... as suggested by @kasei in https://github.com/w3c/sparql-query/issues/95#issuecomment-1584615608

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/98.html" title="Last updated on Jun 22, 2023, 5:40 PM UTC (8e30456)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/98/095c3a0...8e30456.html" title="Last updated on Jun 22, 2023, 5:40 PM UTC (8e30456)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -8770,7 +8770,7 @@ Let E := [], a list of pairs of the form (variable, expression)
 
 If Q contains GROUP BY exprlist
    Let G := Group(exprlist, ToList(P))
-Else If Q contains an aggregate in <code>SELECT</code>, <code>HAVING</code>, <code>ORDER BY</code>
+Else If Q contains an aggregate in SELECT, HAVING, ORDER BY
    Let G := Group((1), ToList(P))
 Else
    skip the rest of the aggregate step
--- a/spec/index.html
+++ b/spec/index.html
@@ -9515,24 +9515,24 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
               ..., agg<sub>n</sub>→val<sub>n</sub> | key in K and key→val<sub>i</sub> in
               S<sub>i</sub> for each 1 &lt;= i &lt;= n }</p>
           </div>
-          <p>Flatten is a function which is used to collapse multisets of lists into a multiset, so
-            for example { (1, 2), (3, 4) } becomes { 1, 2, 3, 4 }.</p>
+          <p>Flatten is a function which is used to collapse a sequence of lists into a single list.
+            For example, [(1,&nbsp;2), (3,&nbsp;4)] becomes (1, 2, 3, 4).</p>
           <div class="defn">
             <p><b>Definition: Flatten</b></p>
-            <p>The Flatten(M) function takes a multiset of lists, M {(L<sub>1</sub>, L<sub>2</sub>,
-              ...), ...}, and returns the multiset { x | L in M and x in L }.</p>
+            <p>The Flatten(S) function takes a sequence of lists, S = [(L<sub>1</sub>, L<sub>2</sub>,
+              ...), ...], and returns the list ( x | L in S and x in L ).</p>
           </div>
           <section id="setFunctions">
             <h5>Set Functions</h5>
             <p>The set functions which underlie SPARQL aggregates all have a common signature:
-              SetFunc(M), or SetFunc(M, scalarvals) where M is a multiset of lists, and scalarvals is
+              SetFunc(S), or SetFunc(S, scalarvals) where S is a sequence of lists, and scalarvals is
               one or more scalar values that are passed to the set function indirectly via the ( ...
               ; key=value ) syntax for aggregates in the SPARQL grammar. The only use of this that is
               supported by the built-in aggregates in SPARQL Query 1.1 is <code>GROUP_CONCAT</code>,
               as in <code>GROUP_CONCAT(?x ; separator=", ")</code>.</p>
             <p>Note that the name "Set Function" is somewhat historical — the arguments to set
-              functions are in fact multisets. The name is retained due to the commonality with SQL
-              Set Functions, which also operate over multisets.</p>
+              functions are in fact sequences. The name is retained due to the commonality with SQL
+              Set Functions, which operate over multisets.</p>
             <p>The set functions defined in this document are Count, Sum, Min, Max, Avg,
               GroupConcat, and Sample — corresponding to the aggregates <code>COUNT</code>,
               <code>SUM</code>, <code>MIN</code>, <code>MAX</code>, <code>AVG</code>,
@@ -9550,10 +9550,10 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
               has a bound, non-error value within the aggregate group.</p>
             <div class="defn">
               <p><b>Definition: <span id="defn_aggCount">Count</span></b></p>
-              <pre class="code nohighlight">xsd:integer Count(multiset M)</pre>
-              <p>N = Flatten(M)</p>
-              <p>remove error elements from N</p>
-              <p>Count(M) = card[N]</p>
+              <pre class="code nohighlight">xsd:integer Count(sequence S)</pre>
+              <p>L = Flatten(S)</p>
+              <p>remove error elements from L</p>
+              <p>Count(S) = card[L]</p>
             </div>
           </section>
           <section id="aggSum">
@@ -9565,13 +9565,14 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
               be 6.0 (float).</p>
             <div class="defn">
               <p><b>Definition: <span id="defn_aggSum">Sum</span></b></p>
-              <pre class="code nohighlight">numeric Sum(multiset M)</pre>
-              <p>Sum(M) = Sum(ToList(Flatten(M))).</p>
-              <p>Sum(S) = op:numeric-add(S<sub>1</sub>, Sum(S<sub>2..n</sub>)) when card[S] &gt;
+              <pre class="code nohighlight">numeric Sum(sequence S)</pre>
+              <p>L = Flatten(S)</p>
+              <p>Sum(S) = Sum(L)</p>
+              <p>Sum(L) = op:numeric-add(L<sub>1</sub>, Sum(L<sub>2..n</sub>)) when card[L] &gt;
                 1<br>
-                Sum(S) = op:numeric-add(S<sub>1</sub>, 0) when card[S] = 1<br>
-                Sum(S) = "0"^^xsd:integer when card[S] = 0</p>
-              <p>In this way, Sum({1, 2, 3}) = op:numeric-add(1, op:numeric-add(2,
+                Sum(L) = op:numeric-add(L<sub>1</sub>, 0) when card[L] = 1<br>
+                Sum(L) = "0"^^xsd:integer when card[L] = 0</p>
+              <p>In this way, Sum( (1, 2, 3) ) = op:numeric-add(1, op:numeric-add(2,
                 op:numeric-add(3, 0))).</p>
             </div>
           </section>
@@ -9581,11 +9582,11 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
             average value for an expression over a group. It is defined in terms of Sum and Count.
             <div class="defn">
               <p><b>Definition: <span id="defn_aggAvg">Avg</span></b></p>
-              <pre class="code nohighlight">numeric Avg(multiset M)</pre>
-              <p>Avg(M) = "0"^^xsd:integer, where Count(M) = 0</p>
-              <p>Avg(M) = Sum(M) / Count(M), where Count(M) &gt; 0</p>
+              <pre class="code nohighlight">numeric Avg(sequence S)</pre>
+              <p>Avg(S) = "0"^^xsd:integer, where Count(S) = 0</p>
+              <p>Avg(S) = Sum(S) / Count(S), where Count(S) &gt; 0</p>
             </div>
-            <p>For example, Avg({1, 2, 3}) = Sum({1, 2, 3})/Count({1, 2, 3}) = 6/3 = 2.</p>
+            <p>For example, Avg([(1), (2), (3)]) = Sum([(1), (2), (3)])/Count([(1), (2), (3)]) = 6/3 = 2.</p>
           </section>
           <section id="aggMin">
             <h5>Min</h5>
@@ -9595,12 +9596,12 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
               arbitrarily typed expressions.</p>
             <div class="defn">
               <p><b>Definition: <span id="defn_aggMin">Min</span></b></p>
-              <pre class="code nohighlight">term Min(multiset M)</pre>
-              <p>Min(M) = Min(ToList(Flatten(M)))</p>
-              <p>Min({}) = error.</p>
-              <p>The flattened multiset of values passed as an argument is converted to a sequence
-                S, this sequence is ordered as per the <code>ORDER BY ASC</code> clause.</p>
-              <p>Min(S) = S<sub>0</sub></p>
+              <pre class="code nohighlight">term Min(sequence S)</pre>
+              <p>L = Flatten(S)</p>
+              <p>Min(S) = Min(L)</p>
+              <p>The flattened list L of values is ordered as per the <code>ORDER BY ASC</code> clause.</p>
+              <p>Min(L) = L<sub>0</sub> if card[L] > 0<br>
+                Min(L) = error if card[L] = 0</p>
             </div>
           </section>
           <section id="aggMax">
@@ -9611,12 +9612,12 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
               arbitrarily typed expressions.</p>
             <div class="defn">
               <p><b>Definition: <span id="defn_aggMax">Max</span></b></p>
-              <pre class="code nohighlight">term Max(multiset M)</pre>
-              <p>Max(M) = Max(ToList(Flatten(M)))</p>
-              <p>Max({}) = error.</p>
-              <p>The multiset of values passed as an argument is converted to a sequence S, this
-                sequence is ordered as per the <code>ORDER BY DESC</code> clause.</p>
-              <p>Max(S) = S<sub>0</sub></p>
+              <pre class="code nohighlight">term Max(sequence S)</pre>
+              <p>L = Flatten(S)</p>
+              <p>Max(S) = Max(L)</p>
+              <p>The flattened list L of values is ordered as per the <code>ORDER BY DESC</code> clause.</p>
+              <p>Max(L) = L<sub>0</sub> if card[L] > 0<br>
+                Max(L) = error if card[L] = 0</p>
             </div>
           </section>
           <section id="aggGroupConcat">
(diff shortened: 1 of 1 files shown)


hartig:
Currently, this change uses Option 1 of https://github.com/w3c/sparql-query/issues/95#issuecomment-1590062135 to convert the input to the [Group operator](https://www.w3.org/TR/sparql11-query/#defn_algGroup) from a multiset of solution mappings into a sequence of solution mappings.

Following this change, this PR changes the definitions of both the [Group operator](https://www.w3.org/TR/sparql11-query/#defn_algGroup) and the [Aggregation operator](https://www.w3.org/TR/sparql11-query/#defn_algAggregation).

In particular, the definition of the Group operator now captures explicitly that the operator produces partial functions from keys to solution sequences.

Similarly, the definition of the Aggregation operator now captures that:
1. such partial functions from keys to solution sequences are the input to this operator,
2. the operator internally produces lists of tuples rather than sets of tuples (see, in particular, the change to the function M),
3. the given "set function" is invoked with such a list of tuples rather than with a set of tuples.

Additionally, I have adapted the example below the definition of the Aggregation operator according to these changes.

However, that's only a part of the changes that are needed to implement @kasei 's proposal. The next step is to adapt the definitions of the set functions. For instance, in [Section 18.5.1.1 Set Functions](https://www.w3.org/TR/sparql-query/#setFunctions), the symbol M must now denote a lists of lists rather than a multiset of lists, and the same holds for M in [Section 18.5.1.2 Count](https://www.w3.org/TR/sparql-query/#aggCount) and also in the next sections. As another consequence, the definition of the Flatten function (as used in [Section 18.5.1.2 Count](https://www.w3.org/TR/sparql-query/#aggCount)) needs to be changed as well.

@kasei any comments so far?

spec/index.html
```diff
@@ -9441,22 +9445,22 @@ <h4>Aggregate Algebra</h4>
...
+              key<sub>m</sub>→Ψ<sub>m</sub> } )<br>
+              &nbsp;&nbsp;&nbsp;= { (key, F(Ψ)) | key → Ψ in { key<sub>1</sub>→Ψ<sub>1</sub>, ...,
+              key<sub>m</sub>→Ψ<sub>m</sub> } }</p>
             <p>where<br>
-              &nbsp;&nbsp;M(Ω) = { ListEval(exprlist, μ) | μ in Ω }<br>
-              &nbsp;&nbsp;F(Ω) = func(M(Ω), scalarvals), for non-DISTINCT<br>
-              &nbsp;&nbsp;F(Ω) = func(Distinct(M(Ω)), scalarvals), for DISTINCT</p>
+              &nbsp;&nbsp;M(Ψ) = [ ListEval(exprlist, μ) | μ in Ψ ]<br>
+              &nbsp;&nbsp;F(Ψ) = func(M(Ψ), scalarvals), for non-DISTINCT<br>
+              &nbsp;&nbsp;F(Ψ) = func(Distinct(M(Ψ)), scalarvals), for DISTINCT</p>
```
kasei, line 9460: We need to define what `Distinct(sequence)` means. Perhaps that for every element `e_i` in `sequence`, there does not exist `e_j` in `sequence` where `e_i = e_j` and `j > I`.

kasei:
Other than the comment about needing to define `Distinct(sequence)`, I think this looks good.

hartig:
I now have changed the definitions of the set functions such that they are defined over sequences of lists, rather than over multisets of lists.

With this change, the PR is ready to be reviewed.

spec/index.html
```diff
@@ -8765,9 +8769,9 @@ <h5>Grouping and Aggregation</h5>
 Let E := [], a list of pairs of the form (variable, expression)
 
 If Q contains GROUP BY exprlist
-   Let G := Group(exprlist, P)
+   Let G := Group(exprlist, ToList(P))
 Else If Q contains an aggregate in SELECT, HAVING, ORDER BY
```
TallTed, line 8773:
```suggestion
Else If Q contains an aggregate in <code>SELECT</code>, <code>HAVING</code>, <code>ORDER BY</code>
```

spec/index.html
hartig: I agree. I don't like this way of defining Distinct either. Instead, your proposal above is much more preferable I think. Yet, I tried to avoid changing anything that is not directly relevant for the issue addressed by this PR. However, I can of course add this change as well to this PR (once I am back on my computer, tomorrow).

TallTed:
(tangent... could be moved to a new issue)

I'm probably in the minority, and it certainly would require a chunk of work by editors and other contributors, but I find myself wishing that a lot more content were `<code>` (or ``` ` ```) tagged, especially all the logical expressions. I find such text a good deal more comprehensible with such typographical adjustments, than with most everything being in the main "body" typeface. I'll cheerfully do a lot if not all of the necessary tag additions, if others are OK with it happening.

gkellogg:
> (tangent... could be moved to a new issue)
> 
> I'm probably in the minority, and it certainly would require a chunk of work by editors and other contributors, but I find myself wishing that a lot more content were `<code>` (or `` ` ``) tagged, especially all the logical expressions. I find such text a good deal more comprehensible with such typographical adjustments, than with most everything being in the main "body" typeface. I'll cheerfully do a lot if not all of the necessary tag additions, if others are OK with it happening.

`var` (or synonym `|`) is also useful markup for things like <var>μ</var> in <var>Ψ</var>

hartig:
I have created two issues related to the two markup suggestions, see #101 and #102.

spec/index.html
hartig: @kasei I have addressed your comment about the definition of Distinct. See commit https://github.com/w3c/sparql-query/pull/98/commits/ff80086d0b37ff19eff77b72b664db42d95f0326

hartig:
I have pushed a new commit to this PR (see https://github.com/w3c/sparql-query/pull/98/commits/d808fb81cb44ad57931f326ab0af0fcca3fd474a). This commit addresses [Andy's concerns about my changes to the  definition of Distinct](https://github.com/w3c/sparql-query/commit/ff80086d0b37ff19eff77b72b664db42d95f0326#commitcomment-118508970) by implementing [his suggestion to separate the definitions](https://github.com/w3c/sparql-query/commit/ff80086d0b37ff19eff77b72b664db42d95f0326#commitcomment-118517879); that is, with this new commit, the definition of Distinct is now back to how it always was and there is a new function called Dedup that is used in the definition of the Aggregation operator. Additionally, this commit addresses the following comments:

[Comment](https://github.com/w3c/sparql-query/commit/ff80086d0b37ff19eff77b72b664db42d95f0326#r118399749) by @kasei 
> This wording seems confusing to me. If there are duplicated elements in Ψ, I would not think of the distinct sequence as "containing every element in Ψ" (since some of them will be omitted because they are duplicates).

Addressed by changing "Every element" to "Every unique element" [as suggested by](https://github.com/w3c/sparql-query/commit/ff80086d0b37ff19eff77b72b664db42d95f0326#r118402147) @TallTed .

[Comment](https://github.com/w3c/sparql-query/commit/ff80086d0b37ff19eff77b72b664db42d95f0326#r118507227) by @afs 
> s/≠/is not the same term as/

I kept the ≠ in the condition for the indexes _i_ and _j_ but integrated "is not the same term as" into the text of the corresponding bullet point.

[Comment](https://github.com/w3c/sparql-query/commit/ff80086d0b37ff19eff77b72b664db42d95f0326#r118508717) by @afs 
> s/every/any/ because the text names the chosen as e1 and e2.

Done.

kasei (approved):

spec/index.html
```diff
@@ -9368,22 +9368,10 @@ <h3>SPARQL Algebra</h3>
...
-                <li><var>i<sub>2</sub></var> is the smallest natural number such that <var>e<sub>2</sub></var> is at the <var>i<sub>2</sub></var>-th position in <var>Ψ</var>,</li>
-                <li><var>j<sub>1</sub></var> is the position of <var>e<sub>1</sub></var> in Distinct(<var>Ψ</var>), and</li>
-                <li><var>j<sub>2</sub></var> is the position of <var>e<sub>2</sub></var> in Distinct(<var>Ψ</var>).</li>
-              </ul>
-            </li>
-          </ol>
+          <p>Let Ψ be a sequence of solution mappings. We define:</p>
+          <p>Distinct(Ψ) = [ μ | μ in Ψ ]</p>
+          <p>card[Distinct(Ψ)](μ) = 1</p>
+          <p>The order of Distinct(Ψ) must preserve any ordering given by OrderBy.</p>
```
```suggestion
          <p>Let <var>Ψ</var> be a sequence of solution mappings. We define:</p>
          <p><code>Distinct(<var>Ψ</var>) = [ <var>μ</var> | <var>μ</var> in <var>Ψ</var> ]</code></p>
          <p><code>card[Distinct(<var>Ψ</var>)](<var>μ</var>) = 1</code></p>
          <p>The order of <code>Distinct(<var>Ψ</var>)</code> must preserve any ordering given by <code>OrderBy</code>.</p>
```

spec/index.html
```diff
@@ -9469,10 +9457,25 @@ <h4>Aggregate Algebra</h4>
             <p>where<br>
               &nbsp;&nbsp;M(Ψ) = [ ListEval(exprlist, μ) | μ in Ψ ]<br>
               &nbsp;&nbsp;F(Ψ) = func(M(Ψ), scalarvals), for non-<code>DISTINCT</code><br>
```
```suggestion
              &nbsp;&nbsp;<code>M(<var>Ψ</var>) = [ ListEval(exprlist, <var>μ</var>) | <var>μ</var> in <var>Ψ</var> ]</code><br>
              &nbsp;&nbsp;<code>F(<var>Ψ</var>) = func(M(<var>Ψ</var>), scalarvals)</code>, for non-<code>DISTINCT</code><br>
```

spec/index.html
```diff
@@ -9469,10 +9457,25 @@ <h4>Aggregate Algebra</h4>
...
+              <li>For any two elements <var>e<sub>1</sub></var> and <var>e<sub>2</sub></var> in Dedup(M(Ψ)), the relative order of their first occurrences in M(Ψ) is preserved in Dedup(M(Ψ)). That is, if <var>i<sub>1</sub></var>&nbsp;&lt;&nbsp;<var>i<sub>2</sub></var>, then <var>j<sub>1</sub></var>&nbsp;&lt;&nbsp;<var>j<sub>2</sub></var>, where
+                <ul>
+                  <li><var>i<sub>1</sub></var> is the smallest natural number such that <var>e<sub>1</sub></var> is at the <var>i<sub>1</sub></var>-th position in M(Ψ),</li>
+                  <li><var>i<sub>2</sub></var> is the smallest natural number such that <var>e<sub>2</sub></var> is at the <var>i<sub>2</sub></var>-th position in M(Ψ),</li>
+                  <li><var>j<sub>1</sub></var> is the position of <var>e<sub>1</sub></var> in Dedup(M(Ψ)), and</li>
+                  <li><var>j<sub>2</sub></var> is the position of <var>e<sub>2</sub></var> in Dedup(M(Ψ)).</li>
```
TallTed, line 9471:
```suggestion
              &nbsp;&nbsp;<code>F(<var>Ψ</var>) = func(Dedup(M(<var>Ψ</var>)), scalarvals)<code>, for <code>DISTINCT</code></p>
            <p>with <code>Dedup(M(<var>Ψ</var>))<code> being an order-preserving, duplicate-free version of the sequence <code>M(<var>Ψ</var>)</code>; that is, <code>Dedup(M(<var>Ψ</var>))</code> is a sequence of RDF terms that has the following four properties.</p>
            <ol>
              <li>Every unique element in <code>M(<var>Ψ</var>)</code> is contained in <code>Dedup(M(<var>Ψ</var>))<code>.</li>
              <li>Every element in <code>Dedup(M(<var>Ψ</var>))</code> is contained in <code>M(<var>Ψ</var>)</code>.</li>
              <li><code>Dedup(M(<var>Ψ</var>))</code> is free of duplicates. That is, the element at the |i|-th position in <code>Dedup(M(<var>Ψ</var>))</code> is not the same term as the element at the |j|-th position in <code>Dedup(M(<var>Ψ</var>))</code> for every two natural numbers |i| and |j| such that |i| &ne; |j|.</li>
              <li>For any two elements <var>e<sub>1</sub></var> and <var>e<sub>2</sub></var> in <code>Dedup(M(<var>Ψ</var>))</code>, the relative order of their first occurrences in <code>M(<var>Ψ</var>)</code> is preserved in <code>Dedup(M(<var>Ψ</var>))</code>. That is, if <var>i<sub>1</sub></var>&nbsp;&lt;&nbsp;<var>i<sub>2</sub></var>, then <var>j<sub>1</sub></var>&nbsp;&lt;&nbsp;<var>j<sub>2</sub></var>, where
                <ul>
                  <li><var>i<sub>1</sub></var> is the smallest natural number such that <var>e<sub>1</sub></var> is at the <var>i<sub>1</sub></var>-th position in <code>M(<var>Ψ</var>)</code>,</li>
                  <li><var>i<sub>2</sub></var> is the smallest natural number such that <var>e<sub>2</sub></var> is at the <var>i<sub>2</sub></var>-th position in <code>M(<var>Ψ</var>)</code>,</li>
                  <li><var>j<sub>1</sub></var> is the position of <var>e<sub>1</sub></var> in <code>Dedup(M(<var>Ψ</var>))</code>, and</li>
                  <li><var>j<sub>2</sub></var> is the position of <var>e<sub>2</sub></var> in <code>Dedup(M(<var>Ψ</var>))</code>.</li>
```

spec/index.html
```diff
@@ -9469,10 +9457,25 @@ <h4>Aggregate Algebra</h4>
...
+                </ul>
+              </li>
+            </ol>
+
             <p><b>Special Case:</b> when <code>COUNT</code> is used with the expression
               <code>*</code> the value of F will be the cardinality of the group solution sequence,
-              <code>card[Ψ]</code>, or <code>card[Distinct(Ψ)]</code> if the <code>DISTINCT</code>
+              <code>card[Ψ]</code>, or <code>card[Dedup(Ψ)]</code> if the <code>DISTINCT</code>
```
TallTed, line 9478:
```suggestion
              <code>card[<var>Ψ</var>]</code>, or <code>card[Dedup(<var>Ψ</var>)]</code> if the <code>DISTINCT</code>
```

spec/index.html
hartig: @TallTed I agree with your suggestions to markup variables in the formalisms by using `<var>` tags (see also #102). I didn't do it in this PR because I wanted to avoid overloading the PR with too many types of changes at the same time.

However, I don't agree with the other type of change in your suggestions here, namely to markup formulas and mathematical functions (such as Dedup, F, and M) by using `<code>` tags. I mean, I'm in favor of using `<code>` tags for language keywords and for AST symbols, but not also for the mathematical symbols and formulas. These are not code that should appear as typewriter/monospaced font.

TallTed: I think we should strive for the best presentation we can achieve.

Is there another tag that should be used? Or another class to apply to such code tags, similar to what @afs mentioned for sparql code highlighting?

Formulas and mathematical functions that are written up in textbooks on these and related subjects are not generally printed using body text (nor, admittedly, monospaced/typewriter) styles. I have hopes we can come close with what we produce.

hartig: @TallTed I am with you. Formulas and snippets thereof (e.g., names of mathematical functions) should be typeset to look different than the regular text. I have created issue #103 as a reminder to implement such a change throughout the whole document.

Now, are you okay with leaving such mark up out of this PR?  I would like to avoid overloading the PR with too many types of changes, in particular because I am worried that we will end up discussing the particular markup to be used, which should better be done in a separate PR.

spec/index.html
```diff
@@ -8765,9 +8769,9 @@ <h5>Grouping and Aggregation</h5>
 Let E := [], a list of pairs of the form (variable, expression)
 
 If Q contains GROUP BY exprlist
-   Let G := Group(exprlist, P)
-Else If Q contains an aggregate in SELECT, HAVING, ORDER BY
-   Let G := Group((1), P)
+   Let G := Group(exprlist, ToList(P))
+Else If Q contains an aggregate in <code>SELECT</code>, <code>HAVING</code>, <code>ORDER BY</code>
+   Let G := Group((1), ToList(P))
```
Tpt, line 8774: It does not seems to cover the "in case of implicit grouping the single group must exist and might be empty" behavior. But it's not a problem for this MR, just noticing this here.

Tpt (approved):
Thank you! I allowed myself some questions I had while reviewing this MR but they are not directly related with it.

spec/index.html
hartig: @TallTed 
>  I confess to flagging things when and where I notice them, and trying to fix them as quickly as possible, which often means first-attempt goes into an inappropriate PR or issue, instead of creating a fresh issue straight-off. I'll continue to try to do better at raising those distinct issues. :-)

No worries! Your reviews are much appreciated.

hartig: I am marking this conversation (included the change suggestion) as resolved. The suggested changes will be picked up again in the context of issues #101, #102, and #103.

spec/index.html
```diff
@@ -8765,9 +8769,9 @@ <h5>Grouping and Aggregation</h5>
 Let E := [], a list of pairs of the form (variable, expression)
 
 If Q contains GROUP BY exprlist
-   Let G := Group(exprlist, P)
-Else If Q contains an aggregate in SELECT, HAVING, ORDER BY
-   Let G := Group((1), P)
+   Let G := Group(exprlist, ToList(P))
+Else If Q contains an aggregate in <code>SELECT</code>, <code>HAVING</code>, <code>ORDER BY</code>
```
hartig, line 8773: Removing the `<code>` elements again from the algorithm as it makes the algorithm look weird. Better come up with a proper approach to using `<code>` tags when addressing issue #101 
```suggestion
Else If Q contains an aggregate in SELECT, HAVING, ORDER BY
```

spec/index.html
```diff
@@ -8745,7 +8745,11 @@ <h5>Grouping and Aggregation</h5>
             <p>Step: GROUP BY</p>
             <p>If the <code>GROUP BY</code> keyword is used, or there is implicit grouping due to the
               use of aggregates in the projection, then grouping is performed by the 
-              <a href="#defn_algGroup">Group</a> function. It divides the solution set into groups of one or
+              <a href="#defn_algGroup">Group</a> function.
+              In this case, before grouping, the solution set is converted into a solution
+              sequence by applying the <a href="#defn_algToList">ToList</a> function.
+              Next, the <a href="#defn_algGroup">Group</a> function
+              divides this solution sequence into groups of one or
```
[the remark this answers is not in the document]
hartig: @Tpt  Your observation is right but, indeed, I would also ignore this edge case here.

spec/index.html
hartig: @Tpt Good catch! However, I think that avoiding to use the Group operator is not possible because the translation algorithm does not know whether `P` produces an empty result. So, instead, the definition of the Group operator must be extended to handle this edge case differently.

Notice also that this problem is independent of the changes in this PR; it is also exists in the current spec text. To avoid overloading the PR, to you mind creating a separate issue?

spec/index.html
```diff
@@ -9415,10 +9431,10 @@ <h4>Aggregate Algebra</h4>
             <div id="defn_algGroup">
               <b>Definition: Group</b>
             </div>
-            <p>Group evaluates a list of expressions against a solution sequence, producing a set
+            <p>Group evaluates a list of expressions against a solution sequence Ψ, producing a set
               of partial functions from keys to solution sequences.</p>
-            <p>Group(exprlist, Ω) = { ListEval(exprlist, μ) → { μ' | μ' in Ω, ListEval(exprlist, μ)
-              = ListEval(exprlist, μ') } | μ in Ω }</p>
+            <p>Group(exprlist, Ψ) = { ListEval(exprlist, μ) → [ μ' | μ' in Ψ, ListEval(exprlist, μ)
+              = ListEval(exprlist, μ') ] | μ in Ψ }</p>
```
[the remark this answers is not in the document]
hartig: @Tpt No, the output is indeed a set of partial functions from "keys" to solution sequences. Notice the curly braces and the "| μ in Ψ" within the curly braces. This creates a set and the elements of this set (i.e., the partial functions) are created by iterating over every solution mapping μ in Ψ. For every such solution mapping, a partial function for the set is created, but some of these partial functions are the same and, thus, are kept only once for the set. The case in which such partial functions are the same is for every equivalence group of solution mappings for which ListEval(exprlist, μ) produces the same "key".

spec/index.html
Tpt: Thank you! I have opened #105.

spec/index.html
```diff
@@ -9415,10 +9431,10 @@ <h4>Aggregate Algebra</h4>
             <div id="defn_algGroup">
               <b>Definition: Group</b>
             </div>
-            <p>Group evaluates a list of expressions against a solution sequence, producing a set
+            <p>Group evaluates a list of expressions against a solution sequence Ψ, producing a set
               of partial functions from keys to solution sequences.</p>
```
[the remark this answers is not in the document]
Tpt: I have opened #106 about it.

hartig:
@afs @rubensworks are you still planning to review this PR?
Otherwise, I would merge it so that we can make progress on a number of related issues (#94 #99 #105 #106 #107, and eventually also #101 #102 #103).

afs:
Yes, I am hoping to review it.

Sorry for the delay.

Reviewing changes in RDF docs has also taken up time.

hartig:
No worries. Take your time! I just wanted to quickly check.

rubensworks (approved):
Apologies for the delay, I lost track of this one.

Issue #99, opened 2023-06-15
Inconsistencies between the definitions of the set functions


hartig:
When working on PR #98 I noticed several inconsistencies between the definitions of the set functions.
1. Some definitions use "card[S]" to denote the cardinality of a sequence S (e.g., [Sec.18.5.1.3 Sum](https://www.w3.org/TR/sparql-query/#aggSum)) whereas others use "|S|" (e.g., [Sec.18.5.1.7 GroupConcat](https://www.w3.org/TR/sparql-query/#aggGroupConcat)). My proposal is to consistently use the latter.
2. Some definitions use the word "when" when capturing conditions (e.g., [Sec.18.5.1.3 Sum](https://www.w3.org/TR/sparql-query/#aggSum)) whereas others use "where" (e.g., [Sec.18.5.1.4 Avg](https://www.w3.org/TR/sparql-query/#aggAvg)). My proposal is to consistently use the word "if".
3. Additionally, the definition of [GroupConcat](https://www.w3.org/TR/sparql-query/#aggGroupConcat) contains only the function signature without the scalarvals argument. A second function signature needs to be added that contains the scalarvals argument.

hartig:
To avoid merge conflicts with the changes in PR #98 I will wait with addressing the points in this issue until we are done with PR #98.

hartig:
Another inconsistency that I forgot to mention:
4. In some definitions, the indexes of elements in the given sequence start from 1 (e.g., [Sec.18.5.1.3 Sum](https://www.w3.org/TR/sparql-query/#aggSum)) whereas other definitions use indexes starting from 0 (e.g., [Sec.18.5.1.7 GroupConcat](https://www.w3.org/TR/sparql-query/#aggGroupConcat)). My proposal is to consistently use the former (i.e., starting from 1).

hartig:
Closed by PR #108

PR #108, opened 2023-06-29
Removes inconsistencies between the definitions of the set functions

hartig:
... as pointed out in #99

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/108.html" title="Last updated on Jun 29, 2023, 10:56 AM UTC (dedb4d0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/108/ae554d5...dedb4d0.html" title="Last updated on Jun 29, 2023, 10:56 AM UTC (dedb4d0)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -9649,11 +9649,13 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
               SEPARATOR.</p>
             <div class="defn">
               <p><b>Definition: <span id="defn_aggGroupConcat">GroupConcat</span></b></p>
-              <pre class="code nohighlight">literal GroupConcat(sequence S)</pre>
-              <p>If the "separator" scalar argument is absent from GROUP_CONCAT then it is taken to
-                be the "space" character, unicode codepoint U+0020.</p>
+              <pre class="code nohighlight">literal GroupConcat(sequence S, function scalarvals)</pre>
+              <p>If the scalarvals argument is absent from GROUP_CONCAT, then scalarvals is taken to
+                be the empty function.</p>
+              <p>|separator| = scalarvals("separator") if scalarvals is defined for the argument "separator"</p>
+              <p>|separator| = the "space" character, unicode codepoint U+0020, if scalarvals is undefined for the argument "separator"</p>
               <p>L = Flatten(S)</p>
-              <p>GroupConcat(S, scalarvals) = GroupConcat(L, scalarvals("separator"))</p>
+              <p>GroupConcat(S, scalarvals) = GroupConcat(L, |separator|)</p>
               <p>GroupConcat(L, sep) = "" if <a href="#defn_Card">Card</a>(L) = 0</p>
               <p>GroupConcat(L, sep) = CONCAT("", L<sub>0</sub>) if
                 <a href="#defn_Card">Card</a>(L) = 1</p>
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -9537,6 +9538,11 @@ <h4>Aggregate Algebra</h4>
             <p>The Flatten(S) function takes a sequence of lists, S = [(L<sub>1</sub>, L<sub>2</sub>,
               ...), ...], and returns the list ( x | L in S and x in L ).</p>
           </div>
+          <p>Card is a function that returns the cardinality of a sequence or a list of elements (which may be solution mappings or other types of elements, depending on the context).
+          <div class="defn">
+            <p><b>Definition: <span id="defn_Card">Card</span></b></p>
+            <p>Given a sequence or a list |L|, Card(|L|) is the cardinality of |L|.</p>
```
rubensworks, line 9544: Should the list be named just `L`, with `|L|` being syntatic sugar for `Card(L)`?

```suggestion
            <p>Given a sequence or a list L, Card(L) is the cardinality of L.</p>
```

kasei: That's [respec shorthand markup for a variable](https://respec.org/docs/#variable-shorthands).

rubensworks: Ah, TIL, thanks for letting me know!

rubensworks (approved):

Tpt (approved):

afs:
`card[ ]` appears in other places as well.

There's a definition of the notation "Write card[Ω](μ) for the cardinality of solution mapping μ in a multiset of mappings Ω." in 18.3

If they are the same, then they need to be aligned.
If they are different and `card(` is specific to aggregation usage, then maybe we need a different way to write it.

hartig:
I am aware of `card[Ω](μ)` as used in other parts of the spec. That function has two arguments---a multiset Ω of solution mappings and a solution mapping μ---and, as you quote, it denotes how often the solution mapping μ occurs within Ω. Before this PR, a similar notation was used also for a different purpose: That is, `card[L]` was used to denote the cardinality (the size) of some list/sequence `L`. But this notation was not defined anywhere in the spec. This PR fixes this issue by explicitly defining a function called Card. Notice also that, in all the places in which I use this function, the function name is linked to the definition. We may do the same for the notation `card[Ω](μ)`.

afs:
`card[Ω]` evaluates to a function that gives the number of instances of `μ`.

We could change that "multiplicity" or something based on that word ("multiplicity" seems a bit long but may be necessary).

hartig:
> We could change that "multiplicity" or something based on that word ("multiplicity" seems a bit long but may be necessary).

I agree. Yet, I think this is a separate question (separate from this PR here). I have created PR #110 with a proposal of how I think we may replace the notation `card[Ω](μ)` by a function called 'multiplicity' that has a definition that can be cross-referenced.

hartig:
@afs Do you agree that the discussion of whether and how the notation `card[Ω](μ)` should be changed is independent of the changes in this PR and may be taken elsewhere instead (e.g., to PR #110 in which I have proposed such a change for `card[Ω](μ)`)?

If you agree, are there any other changes in this PR that you see as potentially problematic?

Otherwise, I may still merge this PR before going on vacation.

afs (approved):

afs:
Yes, I agree. Merging before your break is a good idea.

Issue #103, opened 2023-06-21
Special typesetting of formulas


hartig:
[As suggested](https://github.com/w3c/sparql-query/pull/98#discussion_r1234664954) by @TallTed, formulas and mathematical functions should be typeset in a way that looks different from the regular text. Quoting @TallTed's comment:

> Is there another tag that should be used? Or another class to apply to such code tags, similar to what @afs mentioned for sparql code highlighting?
>
> Formulas and mathematical functions that are written up in textbooks on these and related subjects are not generally printed using body text (nor, admittedly, monospaced/typewriter) styles. I have hopes we can come close with what we produce.

Related to this issue, and captured as #102, is that the symbols used for variables in such formulas should be marked up using `<var>` (or ``` | ```) .

hartig:
Here is my take on the following questions.

> Is there another tag that should be used? Or another class to apply to such code tags, similar to what @afs mentioned for sparql code highlighting?

I would not use `<code>` tags because formulas and mathematical functions are not code. As there does not seem to exist a special HTML tag for marking up math content, I would simply use `<span>` with a specific CSS class that renders the content to look slightly different from the regular text.

gkellogg:
Have you considered using the `math` Elment the ReSpec? https://developer.mozilla.org/en-US/docs/Web/MathML/Element/math#

TallTed:
[`var`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/var) definitely helps.

At first blush, `<math>` looks like the right direction from there, but I'm concerned that this may require a lot more complex MathML markup. Does anyone here have experience using MathML in such a simplistic way as I think we're contemplating? Is it really as simple as adding a leading `<math display="block">` or `<math display="inline">`, and a trailing `</math>`?

afs:
Could we have some concrete displaying HTML please? (not sure what github tech allows easy raw HTML without interference).

Tpt:
Sadly Github does not seems to support Mathml in comments. Here is [a getting started on MDN](https://developer.mozilla.org/en-US/docs/Learn/MathML/First_steps/Getting_started).

TallTed:
Ugh. The compatibility table claims Chrome 109+ support `<math>`, but the sample displays as `1 3` in my `Version 114.0.5735.133 (Official Build) (x86_64)`. I think that means `<math>` is a lose.

hartig:
[...]

Issue #102, opened 2023-06-16
Mark up math variables using <var>


hartig:
[As suggested](https://github.com/w3c/sparql-query/pull/98#issuecomment-1593740401) by @gkellogg, symbols used in formal definitions (e.g., _μ_ and _Ψ_) should be marked up using `<var>` (or ``` | ```) throughout the whole document.

hartig:
I want to suggest to postpone taking action on this issue here until #105 #106 #107 and #94 are addressed. Each of these issues touches on parts that contain formulas and, thus, PRs for these issues would likely cause merge conflicts with a PR for the issue here.

afs:
And like #101, please style the `<var>` tag when used. It makes detailed CSS control easier.

The SPARQL docs, especially query, have detailed semantic markup (due to EricP) which has been useful.

TallTed:
> [@afs] And like #101, please style the <var> tag when used. It makes detailed CSS control easier.

Explicit advice/instruction is helpful. 

In #101, you suggested `<code class="syntax">`, which I think should only be used for "SPARQL expressions and SPARQL keywords"?  Is that correct?

Presuming so, what would you suggest for the `<var>` `class`?  Is one such `class` all we need for all `<var>` instances?

afs:
The request is to follow the patterns in the existing content. It makes it easier to work on a large document with text tools when there is a consistent approach.

A single class, for now, is better than no class. Having an indirection between reSpec defaults and our docs seems a wise investment.

A better process is that someone should coordinate a _proposal_ for the overall style, presenting alternatives, and examples and allowing others to take that proposal and try out their ideas. By doing that outside of point-by-point pull requests, we can avoid the extra work that can be caused by merge conflicts (see https://github.com/w3c/sparql-query/issues/102#issuecomment-1612479794).

When there is agreement, make the changes in a short period of time.
Let's learn from the situation we have with display on mobile devices.

TallTed:
> follow the patterns in the existing content

Sure, happy to start with that. 

But right now, so far as I can tell, there is no `<var class="...">`, there are only `<var>`, so following the patterns in the existing content means _no class on `var` elements_ which does not seem to be what you wanted.

Which is why I asked —
> what would you suggest for the `<var>` `class`? Is one such `class` all we need for all `<var>` instances?

afs:
> But right now, so far as I can tell, there is no <var class="...">, there are only <var>,

You mean your suggested changes?
[...]

PR #110, opened 2023-07-03
introduces a function called multiplicity to replace card[Ω](μ)

hartig:
In the context of [discussing the new Card(..) function introduced](https://github.com/w3c/sparql-query/pull/108#issuecomment-1618880551) in PR #108, @afs [writes](https://github.com/w3c/sparql-query/pull/108#issuecomment-1618937017):

> `card[Ω]` evaluates to a function that gives the number of instances of `μ`.
>
> We could change that "multiplicity" or something based on that word ("multiplicity" seems a bit long but may be necessary).

This PR here is a proposal to address this comment. In particular, this PR introduces a function called 'multiplicity' with a definition that can be cross-referenced, and (currently) one place where this function is used (namely, in the definition of Filter). The latter is to demonstrate how it would look like to use this functions instead of the notation `card[Ω](μ)`. My proposal is to replace all occurrences of `card[Ω](μ)` in this way.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/110.html" title="Last updated on Oct 16, 2023, 5:33 PM UTC (376e0c6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/110/fc0566b...376e0c6.html" title="Last updated on Oct 16, 2023, 5:33 PM UTC (376e0c6)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -8947,9 +8947,9 @@ The set PV is used later for projection.
         <p>Write <var>μ</var> for solution mappings.</p>
         <p>Write <var>μ<sub>0</sub></var> for the mapping such that <var>dom</var>(<var>μ<sub>0</sub></var>) is the empty set.</p>
         <p>Write <var>Ω<sub>0</sub></var> for the multiset consisting of exactly the empty mapping
-          μ<sub>0,</sub> with multiplicity 1. This is the join identity.</p>
-        <p>Write μ(x) for the solution mapping variable x to RDF term t : { (x, t) }</p>
-        <p>Write Ω(x) for the multiset consisting of exactly μ(?x-&gt;t), that is, { { (x, t) } }
+          <var>μ<sub>0</sub></var>, with multiplicity 1. This is the join identity.</p>
+        <p>Write <var>μ</var>(<var>x</var>) for the solution mapping variable <var>x</var> to RDF term <var>t</var> : { (<var>x</var>, <var>t</var>) }.</p>
+        <p>Write <var>Ω</var>(<var>x</var>) for the multiset consisting of exactly <var>μ</var>(<var>?x</var>-&gt;<var>t</var>), that is, <code>{ { (x, t) } }</code>
           with multiplicity 1.</p>
         <div class="defn">
           <p><b>Definition: <span id="defn_algCompatibleMapping">Compatible Mappings</span></b></p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -8964,12 +8964,12 @@ The set PV is used later for projection.
           μ<sub>2</sub></p>
         <div class="defn">
           <p><b>Definition: <span id="defn_Multiplicity">Multiplicity</span></b></p>
-          <p>Given a multiset&nbsp;Ω of solution mappings and a solution
-            mapping&nbsp;μ, we write multiplicity<sup>Ω</sup>(μ) to denote
-            the number of times μ appears in Ω.</p>
-          <p>Similarly, given a solution sequence&nbsp;Ψ and a solution
-            mapping&nbsp;μ, we write multiplicity<sup>Ψ</sup>(μ) to denote
-            the number of times μ appears in Ψ.</p>
+          <p>Given a multiset&nbsp;<var>Ω</var> of solution mappings and a solution
+            mapping&nbsp;<var>μ</var>, we write <var>multiplicity<sup><var>Ω</var></sup></var>(<var>μ</var>)
+            to denote the number of times <var>μ</var> appears in <var>Ω</var>.</p>
+          <p>Similarly, given a solution sequence&nbsp;<var>Ψ</var> and a solution
+            mapping&nbsp;<var>μ</var>, we write <var>multiplicity<sup><var>Ψ</var></sup></var>(<var>μ</var>)
+            to denote the number of times <var>μ</var> appears in <var>Ψ</var>.</p>
         </div>
         <section id="BGPsparql">
           <h4>SPARQL Basic Graph Pattern Matching</h4>
--- a/spec/index.html
+++ b/spec/index.html
@@ -8944,9 +8944,9 @@ The set PV is used later for projection.
           unordered collection of elements in which each element may appear more than once. It is
           described by a set of elements and a function giving the multiplicity of each of these
           elements (i.e., the number of times the element is contained in the multiset).</p>
-        <p>Write μ for solution mappings.</p>
-        <p>Write μ<sub>0</sub> for the mapping such that dom(μ<sub>0</sub>) is the empty set.</p>
-        <p>Write Ω<sub>0</sub> for the multiset consisting of exactly the empty mapping
+        <p>Write <var>μ</var> for solution mappings.</p>
+        <p>Write <var>μ<sub>0</sub></var> for the mapping such that <var>dom</var>(<var>μ<sub>0</sub></var>) is the empty set.</p>
+        <p>Write <var>Ω<sub>0</sub></var> for the multiset consisting of exactly the empty mapping
           μ<sub>0,</sub> with multiplicity 1. This is the join identity.</p>
         <p>Write μ(x) for the solution mapping variable x to RDF term t : { (x, t) }</p>
         <p>Write Ω(x) for the multiset consisting of exactly μ(?x-&gt;t), that is, { { (x, t) } }
--- a/spec/index.html
+++ b/spec/index.html
@@ -8993,8 +8993,8 @@ The set PV is used later for projection.
             <p>μ is a <b>solution</b> for BGP from G when there is a pattern instance mapping P such
               that P(BGP) is a subgraph of G and μ is the restriction of P to the query variables in
               BGP.</p>
-            <p><a href="#defn_Multiplicity">multiplicity</a><sup>Ω</sup>(μ) = number of distinct RDF instance mappings, σ, such that P = μ(σ)
-              is a pattern instance mapping and P(BGP) is a subgraph of G</p>
+            <p>card[Ω](μ) = card[Ω](number of distinct RDF instance mappings, σ, such that P = μ(σ)
+              is a pattern instance mapping and P(BGP) is a subgraph of G).</p>
           </div>
           <p>If a basic graph pattern is the empty set, then the solution is Ω<sub>0</sub>.</p>
         </section>
--- a/spec/index.html
+++ b/spec/index.html
@@ -8965,10 +8965,10 @@ The set PV is used later for projection.
         <div class="defn">
           <p><b>Definition: <span id="defn_Multiplicity">Multiplicity</span></b></p>
           <p>Given a multiset&nbsp;<var>Ω</var> of solution mappings and a solution
-            mapping&nbsp;<var>μ</var>, we write <var>multiplicity<sup><var>Ω</var></sup></var>(<var>μ</var>)
+            mapping&nbsp;<var>μ</var>, we write <var>multiplicity</var>(<var>μ</var>&nbsp;|&nbsp;<var>Ω</var>)
             to denote the number of times <var>μ</var> appears in <var>Ω</var>.</p>
           <p>Similarly, given a solution sequence&nbsp;<var>Ψ</var> and a solution
-            mapping&nbsp;<var>μ</var>, we write <var>multiplicity<sup><var>Ψ</var></sup></var>(<var>μ</var>)
+            mapping&nbsp;<var>μ</var>, we write <var>multiplicity</var>(<var>μ</var>&nbsp;|&nbsp;<var>Ψ</var>)
             to denote the number of times <var>μ</var> appears in <var>Ψ</var>.</p>
         </div>
         <section id="BGPsparql">
@@ -9262,8 +9262,8 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
           <p>Let Ω be a multiset of solution mappings and expr be an expression. We define:</p>
           <p>Filter(expr, Ω) = { μ | μ in Ω and expr(μ) is an expression that has an
             effective boolean value of true }</p>
-          <p><a href="#defn_Multiplicity">multiplicity</a><sup>Filter(expr, Ω)</sup>(μ)
-            = <a href="#defn_Multiplicity">multiplicity</a><sup>Ω</sup>(μ)</p>
+          <p><a href="#defn_Multiplicity">multiplicity</a>( μ | Filter(expr, Ω) )
+            = <a href="#defn_Multiplicity">multiplicity</a>( μ | Ω )</p>
           <blockquote>
             Note that evaluating an <code>exists(pattern)</code> expression uses the dataset and
             active graph, D(G). See the <a href="#defn_evalFilter">evaluation of filter</a>.
--- a/spec/index.html
+++ b/spec/index.html
@@ -3916,8 +3916,8 @@ SELECT DISTINCT ?name WHERE { ?x foaf:name ?name }
             <h4>REDUCED</h4>
             <p>While the <code>DISTINCT</code> modifier ensures that duplicate solutions are
               eliminated from the solution set, <code>REDUCED</code> simply permits them to be
-              eliminated. The cardinality of any set of variable bindings in a <code>REDUCED</code>
-              solution set is at least one and not more than the cardinality of the solution set with
+              eliminated. The multiplicity of any solution in a <code>REDUCED</code>
+              solution set is at least one and not more than the multiplicity of the solution within the solution set with
               no <code>DISTINCT</code> or <code>REDUCED</code> modifier. For example, using the data
               above, the query</p>
             <div class="queryGroup">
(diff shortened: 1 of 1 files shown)


hartig:
@afs @kasei @Tpt @rubensworks This PR has been around for a while without attracting any discussion. Perhaps that is because I marked the PR as WIP? The reason why I did so is because the PR is not (yet) complete regarding the changes implied by the proposal that I am making here. I just wanted to first get your initial opinions on the proposal before I spend the time implementing the rest of the changes that would follow from the proposal. Therefore, can you please take a look at the [preview of this PR](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/110.html) and let me know what you think. The current changes are in particular in [Section 18.3](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/110.html#BasicGraphPattern) and in the definition of the [Filter operator](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/110.html#defn_algFilter) (as an example to illustrate how the new multiplicity function would be used in the definitions of the algebra operators).

Tpt:
It looks great to me! Using "multiplicity<sup>Ω</sup>(μ)" confused me a bit during a first reading because I read the function name as just "<sup>Ω</sup>(μ)" and "multiplicity" as a sentence word. What about finding a way with typography (italics?) to make clear that "multiplicity" is here part of the function name and not part of the sentence?

hartig:
> It looks great to me! Using "multiplicity<sup>Ω</sup>(μ)" confused me a bit during a first reading because I read the function name as just "<sup>Ω</sup>(μ)" and "multiplicity" as a sentence word. What about finding a way with typography (italics?) to make clear that "multiplicity" is here part of the function name and not part of the sentence?

Good point. I have pushed commit https://github.com/w3c/sparql-query/pull/110/commits/a98dac6938c57665da2a983b58587df95da6452e which uses the feature of ReSpec to explicitly markup symbols in math formulas. Now the function name is rendered in italic font. Additionally, if you look at the HTML locally (it doesn't work in the preview), you can even click on every symbol within the definition to have other occurrences of the symbol highlighted within the definition.

spec/index.html
```diff
@@ -8942,15 +8942,15 @@ <h3>Basic Graph Patterns</h3>
           <a href="https://en.wikipedia.org/w/index.php?title=Multiset">multiset</a>, 
           also known as a <i>bag</i>. A multiset is an
           unordered collection of elements in which each element may appear more than once. It is
-          described by a set of elements and a cardinality function giving the number of occurrences of
-          each element from the set in the multiset.</p>
+          described by a set of elements and a function giving the multiplicity of each of these
+          elements (i.e., the number of times the element is contained in the multiset).</p>
         <p>Write μ for solution mappings.</p>
         <p>Write μ<sub>0</sub> for the mapping such that dom(μ<sub>0</sub>) is the empty set.</p>
         <p>Write Ω<sub>0</sub> for the multiset consisting of exactly the empty mapping
```
TallTed, line 8949:
```suggestion
        <p>Write <var>μ</var> for solution mappings.</p>
        <p>Write <var>μ<sub><var>0</var></sub></var> for the mapping such that <var>dom(<var>μ<sub><var>0</var></sub></var>)</var> is the empty set.</p>
        <p>Write <var>Ω<sub><var>0</var></sub></var> for the multiset consisting of exactly the empty mapping
```

spec/index.html
```diff
@@ -8942,15 +8942,15 @@ <h3>Basic Graph Patterns</h3>
...
-          μ<sub>0,</sub> with cardinality 1. This is the join identity.</p>
+          μ<sub>0,</sub> with multiplicity 1. This is the join identity.</p>
         <p>Write μ(x) for the solution mapping variable x to RDF term t : { (x, t) }</p>
         <p>Write Ω(x) for the multiset consisting of exactly μ(?x-&gt;t), that is, { { (x, t) } }
```
TallTed, line 8952:
```suggestion
          <var>μ<sub><var>0</var></sub></var>, with multiplicity 1. This is the join identity.</p>
        <p>Write <var>μ(<var>x</var>)</var> for the solution mapping variable <var>x</var> to RDF term <var>t</var> : { (<var>x</var>, <var>t</var>) }.</p>
        <p>Write <var>Ω(<var>x</var>)</var> for the multiset consisting of exactly <var>μ(<var>?x</var>-&gt;<var>t</var>)</var>, that is, <code>{ { (x, t) } }</code>
```

TallTed (changes requested):
Should the `<var>` magic not be applied throughout, rather than being confined to the definition?

hartig:
> Should the `<var>` magic not be applied throughout, rather than being confined to the definition?

Yes it should. However, I see that as something that is orthogonal to the proposal in this PR. Instead, I see it as something that we should do systematically across the whole document, as per issues #102 and #103.

TallTed:
> I see that as something that is orthogonal to the proposal in this PR

Given that `<var>` has been applied to the definition, it's not being *not* applied anywhere, so it seems reasonable to apply it to all the text wrapped up in this PR, i.e., the discussion as well as the definition of the new `multiplicity` function.

If it must be done in separate PR(s), can you confirm that I applied it correctly to the complex expressions? I'm not sure I understand how things nest with it.

spec/index.html
hartig: Some comments about your change suggestions here (although, again, I see these things as something that belongs in a PR related to #102 and #103):
* Regarding your proposed `<var>μ<sub><var>0</var></sub></var>`, this should be `<var>μ<sub>0</sub></var>` because `0` is a constant and not the name of a symbol used in the formula.
* Same thing for `<var>Ω<sub><var>0</var></sub></var>`, it should be only `<var>Ω<sub>0</sub></var>`
* Regarding your proposed `<var>dom(<var>μ<sub><var>0</var></sub></var>)</var>`, this should be `<var>dom</var>(<var>μ<sub>0</sub></var>)` -- i.e., I don't think the whole formula should be marked up as a symbol but, instead, the name of the `dom` function is a symbol in itself (and `μ<sub>0</sub>` is another symbol).

spec/index.html
hartig: Similar issues for this change suggestion as for the other one. That is, `<var>...</var>` should not be used to mark up whole formulas but to mark up the individual symbols within the formulas.

hartig:
> If it must be done in separate PR(s), can you confirm that I applied it correctly to the complex expressions? I'm not sure I understand how things nest with it.

I was just working on replies to your two change suggestions ;-)  These replies should be visible now.

spec/index.html
afs: Presumably `<var>` shouldn't nest.

afs:
I agree that changing "cardinality" to "multiplicity" is an improvement.

It is difficult to have helpful styling here. Using the superscripts is one way but text does become small which may be an accessibility issue.

Choosing the presentation should not stop this PR because we can change the presentation later.

<code>multiplicity<sup>LeftJoin(Ω1, Ω2, expr)</sup>(μ) =
&nbsp;&nbsp;&nbsp;&nbsp;multiplicity<sup>Filter(expr, Join(Ω1, Ω2))</sup>(μ) + multiplicity<sup>Diff(Ω1, Ω2, expr)</sup>(μ)
</code>

There is more significant text in the the superscripts.

spec/index.html
hartig: > [@afs] Presumably `<var>` shouldn't nest.

I think there are cases where it should nest. For instance, consider a sequence of solution mappings and we want to write a definition that iterates over these sequence. Then, we may want to use a symbol (say, _i_) to denote the current iteration step and a symbol to denote the _i_-th solution mapping. That latter symbol may contain the symbol _i_ as a part of it. In this case, I would write `<var>μ<sub><var>i</var></sub></var>`.

hartig:
@afs 
> I agree that changing "cardinality" to "multiplicity" is an improvement.

That's a good start ;-)

> It is difficult to have helpful styling here. Using the superscripts is one way but text does become small which may be an accessibility issue.

Yes, I can see this being a problem, in particular because, as you say, the text in the superscripts can be quite significant.

> Choosing the presentation should not stop this PR because we can change the presentation later.

Given that my proposed change needs to be applied to quite a number of definitions (all in [Section 18.5 SPARQL Algebra](https://www.w3.org/TR/sparql12-query/#sparqlAlgebra)), and considering that switching from my currently proposed superscript-based presentation to some other presentation may not be easily done with a search-and-replace, I would prefer we agree on a presentation before I implement the rest of this PR.

So, here is an alternative to the superscript-based presentation: How's about we write _multiplicity_( _μ_ | _Ω_ ) to denote the multiplicity of _μ_ in _Ω_? With this presentation, your example for LeftJoin would then be:

_multiplicity_( _μ_ | LeftJoin(_Ω1_, _Ω2_, _expr_) ) =
    _multiplicity_( _μ_ | Filter(_expr_, Join(_Ω1_, _Ω2_)) )  + _multiplicity_( _μ_ | Diff(_Ω1_, _Ω2_, _expr_) )

Is that better?

afs:
> multiplicity( μ | LeftJoin(Ω1, Ω2, expr) ) =
> multiplicity( μ | Filter(expr, Join(Ω1, Ω2)) ) + multiplicity( μ | Diff(Ω1, Ω2, expr) )
> 
> Is that better?

Yes.

I came back to this PR to suggest using a word `multiplicity( μ in LeftJoin(Ω1, Ω2, expr) )`. Your suggestion of `|` is equally good.

hartig:
I have implemented the change to the presentation of the _multiplicity_ function, using `|` as the separator (commit https://github.com/w3c/sparql-query/pull/110/commits/f1ae8c4f1e83e34f4ec678cf766be17f8061073e).

I am planning to extend this PR to apply the proposal to all relevant parts of the document.

Tpt:
Thank you! `|` is much better. Nitpicking: I find it still a bit confusing because it is often a shortcut to "such that" like in $\{ x | x \in \mu \}$ and it's not the case here. `in` sounds slightly clearer to me. But it's nitpicking, `|` is very fine.

hartig:
This PR is now ready for the actual review.

I have extended it by applying the proposal to all relevant parts of the document, and I have also switched to the correct terminology (cardinality -> multiplicity) in all the relevant places (see commit https://github.com/w3c/sparql-query/pull/110/commits/7e6e6f996745de3914ea69ab5a3d71eb559ce762).

There are still a few mentions of "cardinality" but these really refer to the notion of cardinality of some set or some sequence, rather than to the multiplicity of an element within a multiset or sequence.

Tpt (approved):

rubensworks (approved):

spec/index.html
TallTed: @hartig — Thank you for your suggested fixes, all of which I have applied. I suffer from not having taken any relevant courses, and am learning some of this as we go.

spec/index.html
TallTed: @hartig @afs — Understanding more now, I think the example `<var>μ<sub><var>i</var></sub></var>` should rather be `<var>μ</var><sub><var>i</var></sub>`, i.e., no nested `<var>`. How these render may change my opinion of which should be used.

I have adjusted my suggested changes, above.

spec/index.html
hartig: > [@TallTed] Thank you for your suggested fixes, all of which I have applied. I suffer from not having taken any relevant courses, and am learning some of this as we go.

No worries! We have several complementary types of expertise and skills in the group, and that's a good thing!

The fixes in your change suggestion look good. I will commit them.

spec/index.html
hartig: > [@TallTed] @hartig @afs — Understanding more now, I think the example `<var>μ<sub><var>i</var></sub></var>` should rather be `<var>μ</var><sub><var>i</var></sub>`, i.e., no nested `<var>`. [...]

No, the nesting is correct in this case. There are two symbols here. For one, there is the symbol that denotes the iteration counter, which is _i_. Hence, we would write `<var>i</var>` for that one. Then, there is the symbol that denotes the _i_-th solution mapping, which is not just _μ_ but _μ_<sub>_i_</sub>. Consequently, the `<var>` tag for this symbol needs to be wrapped around the whole thing (i.e., around all of _μ_<sub>_i_</sub>). Yet, since some part of this second symbol happens to be the first symbol (i.e., _i_), this first symbols also needs to be marked up inside the markup of the second symbol. That's how the HTML/Respec representation of the second symbol becomes nested; `<var>μ<sub><var>i</var></sub></var>`.

Regarding your adjusted change suggestion, the first change in it is still incorrect. Instead of `<var>μ</var><sub><var>0</var></sub>`, it should be `<var>μ<sub>0</sub></var>`, because i) `0` is a constant and not the name of a symbol used in the formula and ii) the symbol in the formula is not just _μ_ but _μ_<sub>0</sub>.

TallTed: Thank you, @hartig.

I understand and have applied `<var>μ<sub>0</sub></var>` in place of `<var>μ</var><sub><var>0</var></sub>` above.

And I think I understand the reason for `<var>μ<sub><var>i</var></sub></var>` instead of my previous thinking of `<var>μ</var><sub><var>i</var></sub>`.

(I do note that `<var>μ<sub>0</sub></var>` and `<var>μ</var><sub><var>0</var></sub>` visually render the same — <var>μ<sub>0</sub></var> and <var>μ</var><sub><var>0</var></sub>, respectively — as do `<var>μ</var><sub><var>i</var></sub>` and `<var>μ<sub>i</sub></var>` — <var>μ</var><sub><var>i</var></sub> and <var>μ<sub>i</sub></var>, respectively — so this pursuit of "correct" markup is a somewhat academic exercise, albeit one which I believe is worth performing.)

hartig: @TallTed 

> (I do note that `<var>μ<sub>0</sub></var>` and `<var>μ</var><sub><var>0</var></sub>` visually render the same — <var>μ<sub>0</sub></var> and <var>μ</var><sub><var>0</var></sub>, respectively — as do `<var>μ</var><sub><var>i</var></sub>` and `<var>μ<sub>i</sub></var>` — <var>μ</var><sub><var>i</var></sub> and <var>μ<sub>i</sub></var>, respectively — so this pursuit of "correct" markup is a somewhat academic exercise, albeit one which I believe is worth performing.)

Yes, they render the same. However, in the case of `<var>μ</var><sub><var>i</var></sub>`, if you click on any other occurrence of _i_, the _i_ inside <var>μ</var><sub><var>i</var></sub> gets highlighted, which makes it easy to spot all the places where the symbol _i_ is used. In contrast, when using only `<var>μ<sub>i</sub></var>`, the symbol _i_ inside would not get highlighted. (Note that this 'clicking on symbols' does not work in the previews of PRs.)

Thanks for updating your change suggestion. I will commit it.

gkellogg: Note that if you add `class="algorithm"` someplace in the path containing the `<var>`, ReSpec provides individual highlighting based on each distinct variable. So, the complete variable should be surrounded by `<var>..</var>`. In this case, that argues for `<var>μ<sub>0</sub></var>`, as `μ<sub>0</sub>` is the variable. Otherwise, it would treat `μ` and `<sub>0</sub>` as separate variables. You can actually put the algorithm class on the body if all equivalent variable names are considered the same. Sadly, this seems to be lost when publishing to /TR, although I've seen things lately that could survive publication.

hartig: That's what I meant by 'clicking on symbols'. It is a really nice feature. It would be a shame if this highlightning disappears when publishing.

gkellogg: There's been a ReSpec issue open on this since 2021: https://github.com/w3c/respec/issues/3765. I believe the issue is that it uses JavaScript to make it work, and this was discouraged in published documents. I'm sure it could be done with just CSS, though.

TallTed:
Now we have an [example rendering issue](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/110.html#BasicGraphPattern) —
Screen Shot 2023-10-16 at 04 19 23 PM
In particular, note the collisions between <i>`t`</i> and `)`, and also (though to a lesser degree) between <i>`µ`</i> and `(` and <i>`Ω`</i> and `(`.

This would be resolved by my previous suggestion to add a space between all `{}[]()` and the non-whitespace characters they abut, or by (much more complicated) adding `<i>` tags to all `{}[]()` that abut a `<var>` or other character that is italicized for whatever reason.

hartig:
> Now we have an example rendering issue [...]

Strange. It looks perfectly fine in my browser (Firefox 118.0.2 on Ubuntu Linux):
image

hartig:
I think the only remaining point in this PR is the rendering issue that @TallTed has in his browser (whereas it looks perfectly fine on my end, in different browsers). Ted, given that this issue is not really related to the actual content of this PR, are you okay if I merge the PR?

afs (approved):

TallTed:
For the record, the rendering issue I posted above is seen in **Chrome** _Version 116.0.5845.187 (Official Build) (x86_64)_ on **macOS Mojave** _version 10.14.6 (18G9323)_. (I'm also adding these details to the comment with that screencap.)

I'm fine with merging this PR as is, and treating the rendering issue as distinct.

hartig:
Thanks @TallTed.

PR #115, opened 2023-07-26
Aggregation definition: states that it returns a single partial function

### LINKED ISSUE #107: Simplify the `Group` operation return type
Currently the `Group` operation return type <q>produces a set of partial functions from keys to solution sequences</q> wih the following definition: `Group(exprlist, Ψ) = { ListEval(exprlist, μ) → [ μ' | μ' in Ψ, ListEval(exprlist, μ) = ListEval(exprlist, μ') ] | μ in Ψ }`.

However, with this definition all `ListEval(exprlist, μ)` for a given `μ` are always associated with the same value. Hence, in the set there is only one partial function for each `ListEval(exprlist, μ)` value. So, it should be possible to simplify this definition to return only a partial function from keys to solution sequences instead of a set of partial function.

<s>I believe that the definition `Group(exprlist, Ψ) = k → [ μ | μ in Ψ, ListEval(exprlist, μ) = k ]` should do the job.</s> Edit: not a good solution, see infra.

Tpt:
Closes #107

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/115.html" title="Last updated on Aug 5, 2023, 6:42 AM UTC (e764bd0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/115/797575a...e764bd0.html" title="Last updated on Aug 5, 2023, 6:42 AM UTC (e764bd0)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -8894,7 +8894,7 @@ For each (X AS Var) in SELECT, each HAVING(X), and each ORDER BY X in Q
       Replace V with Sample(V)
       End
   For each aggregate R(args ; scalarvals) now in X
-      # note scalarvals may be omitted, then it's equivalent to the empty set
+      # note scalarvals may be omitted, then it's equivalent to the empty function
       A<span><sub>i</sub></span> := Aggregation(args, R, scalarvals, Grp)
       Replace R(...) with agg<span><sub>i</sub></span> in Q
       i := i + 1
@@ -9555,9 +9555,9 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
               <b>Definition: Aggregation</b>
             </div>
             <p>Let <i>exprlist</i> be a list of expressions or *, <i>func</i> a set function,
-              <i>scalarvals</i> a set of partial functions (possibly empty) passed from the aggregate
+              <i>scalarvals</i> a partial function (possibly empty) passed from the aggregate
               in the query, and let { key<sub>1</sub>→Ψ<sub>1</sub>, ...,
-              key<sub>m</sub>→Ψ<sub>m</sub> } be a set of partial functions from keys to
+              key<sub>m</sub>→Ψ<sub>m</sub> } be a partial function from keys to
               solution sequences as produced by the grouping step.</p>
             <p>Aggregation applies the set function func to the given set and produces a
               single value for each key and group of solutions for that key.</p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -8894,7 +8894,7 @@ For each (X AS Var) in SELECT, each HAVING(X), and each ORDER BY X in Q
       Replace V with Sample(V)
       End
   For each aggregate R(args ; scalarvals) now in X
-      # note scalarvals may be omitted, then it's equivalent to the empty function
+      # note: scalarvals may be omitted; if so, it's equivalent to the empty function
       A<span><sub>i</sub></span> := Aggregation(args, R, scalarvals, Grp)
       Replace R(...) with agg<span><sub>i</sub></span> in Q
       i := i + 1
@@ -9554,13 +9554,13 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
             <div id="defn_algAggregation">
               <b>Definition: Aggregation</b>
             </div>
-            <p>Let <i>exprlist</i> be a list of expressions or *, <i>func</i> a set function,
-              <i>scalarvals</i> a partial function (possibly with an empty domain) passed from the aggregate
-              in the query, and let { key<sub>1</sub>→Ψ<sub>1</sub>, ...,
-              key<sub>m</sub>→Ψ<sub>m</sub> } be a partial function from keys to
+            <p>Let <i>exprlist</i> be a list of expressions or `*`; <i>func</i>, a set function;
+              <i>scalarvals</i>, a partial function (possibly with an empty domain) passed from the aggregate
+              in the query; and { key<sub>1</sub>→Ψ<sub>1</sub>, ...,
+              key<sub>m</sub>→Ψ<sub>m</sub> }, a partial function from keys to
               solution sequences as produced by the grouping step.</p>
-            <p>Aggregation applies the set function func to the given set and produces a
-              single value for each key and group of solutions for that key.</p>
+            <p>Aggregation applies the set function `func` to the given set and produces a
+              single value for each key and a group of solutions for that key.</p>
             <p>Aggregation(exprlist, func, scalarvals, { key<sub>1</sub>→Ψ<sub>1</sub>, ...,
               key<sub>m</sub>→Ψ<sub>m</sub> } )<br>
               &nbsp;&nbsp;&nbsp;= { (key, F(Ψ)) | key → Ψ in { key<sub>1</sub>→Ψ<sub>1</sub>, ...,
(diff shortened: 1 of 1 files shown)


TallTed:
This *changes* the aggregation definition to state that it returns a single partial function.

This does not seem like an editorial change, and this PR does not point back to an issue which is resolved by this change.

Tell us more?

Tpt:
> this PR does not point back to an issue which is resolved by this change.

It's in the description: #107 (the most relevant part is the end of @hartig's comment). Sorry for not highlighting this more.

> This does not seem like an editorial change
> This changes the aggregation definition to state that it returns a single partial function.

That's a great point. The current spec states that `Aggregate` returns a set of partial function each one having a singleton [domain](https://en.wikipedia.org/wiki/Domain_of_a_function) (i.e. a separate function for each key). This is equivalent to returning a single partial function which domain is all keys returned by the aggregation. And the later is imho a bit simpler and more understandable. Hence this PR.
I considered it to be a editorial change because it does not change anything on how the aggregation is evaluated. But feel free to switch it so substantive if you feel so.

afs:
I agree this is "editorial". I don't believe it will affect an implementation - no change to tests and no change to an implementation that reported for SPARQL 1.1 (to avoid the problem of saying "any implementation").

afs:
I looked for other places that may need aligning if `scalarvals` is changing:
The main item is saying what "empty function" is.


I may have missed some - I looked for "scalarvals" and "Aggregation(".

Section "18.2.4.1 Grouping and Aggregation"

The comment says 

> \# note scalarvals may be omitted, then it's equivalent to the empty set

It's not the "empty set" any more.

Section "18.5.1 Aggregate Algebra" (just after the defn box):

> For example, the aggregate expression GROUP_CONCAT(?x ; separator="|") has a scalarvals argument of { "separator" → "|" }.

Is the notation `{ "separator" → "|" }` clear enough? 
I'm happy to keep it as it is and I don't know of an alternative.

The PR change is 

`{ { a -> 1 } , {b -> 2} , { c-> 3} }` becomes `{ a -> 1 , b -> 2 , c-> 3 }`

I haven't found an example in the doc that has multiple mappings.

Section "18.5.1.7 GroupConcat":

> then scalarvals is taken to be the empty function.

No change but do we need to say what "empty function" is in the definition in 18.5.1, or immediately after it?

The example:

> For example, GroupConcat([("a"), ("b"), ("c")], {"separator" → "."}) = "a.b.c".

looks OK - same notation as earlier.

TallTed:
> [@Tpt] The current spec states that Aggregate returns a set of partial function each one having a singleton [domain](https://en.wikipedia.org/wiki/Domain_of_a_function) (i.e. a separate function for each key). This is equivalent to returning a single partial function which domain is all keys returned by the aggregation.

That helps me. I didn't understand that the new language was effectively equivalent to the old language.

> [@afs] I don't believe it will affect an implementation - no change to tests and no change to an implementation that reported for SPARQL 1.1

And that's the key. Given that the above is so, it doesn't matter to me whether the spec change is actually considered editorial or normative; I'll still be OK with the change.

Tpt:
@afs Thank you!

> > \# note scalarvals may be omitted, then it's equivalent to the empty set
> It's not the "empty set" any more.

I have replaced this word by "empty function". Thanks!

> Is the notation `{ "separator" → "|" }` clear enough? 
> I'm happy to keep it as it is and I don't know of an alternative.

I agree. It's the one we use in the spec for all "mathematical-like" fonctions including SPARQL solution mappings. And it make the "empty function" definition quite simple (it's just `{}`).

>  The main item is saying what "empty function" is.

Indeed. 

An option for that might be to rephrase in `#defn_algAggregation` <q><i>scalarvals</i> a partial function (possibly empty)</q> to something like <q><i>scalarvals</i> a partial function (possibly empty i.e. `{}`, the function defined on the empty domain)</q> but I am not sure if it is much more helpful. Do you have a better idea?

afs:
> (possibly empty i.e. {}, the function defined on the empty domain)

Are we going ahead with this change? Looks better to me.

afs (approved):
One non-blocking comment.

spec/index.html
```diff
@@ -9555,9 +9555,9 @@ <h4>Aggregate Algebra</h4>
               <b>Definition: Aggregation</b>
             </div>
             <p>Let <i>exprlist</i> be a list of expressions or *, <i>func</i> a set function,
-              <i>scalarvals</i> a set of partial functions (possibly empty) passed from the aggregate
+              <i>scalarvals</i> a partial function (possibly empty) passed from the aggregate
```
hartig, line 9558:
```suggestion
              <i>scalarvals</i> a partial function (possibly with an empty domain) passed from the aggregate
```

hartig (approved):

Tpt:
>> (possibly empty i.e. {}, the function defined on the empty domain)
> Are we going ahead with this change? Looks better to me.

I have added @hartig suggestion that is fairly similar. Is it ok with you?

spec/index.html
```diff
@@ -8894,7 +8894,7 @@ <h5>Grouping and Aggregation</h5>
       Replace V with Sample(V)
       End
   For each aggregate R(args ; scalarvals) now in X
-      # note scalarvals may be omitted, then it's equivalent to the empty set
+      # note scalarvals may be omitted, then it's equivalent to the empty function
```
TallTed, line 8897:
```suggestion
      # note: scalarvals may be omitted; if so, it's equivalent to the empty function
```

spec/index.html
```diff
@@ -9555,9 +9555,9 @@ <h4>Aggregate Algebra</h4>
               <b>Definition: Aggregation</b>
             </div>
             <p>Let <i>exprlist</i> be a list of expressions or *, <i>func</i> a set function,
```
TallTed, line 9557:
```suggestion
            <p>Let <i>exprlist</i> be a list of expressions or `*`; <i>func</i>, a set function;
```

spec/index.html
```diff
@@ -9555,9 +9555,9 @@ <h4>Aggregate Algebra</h4>
...
-              <i>scalarvals</i> a set of partial functions (possibly empty) passed from the aggregate
+              <i>scalarvals</i> a partial function (possibly with an empty domain) passed from the aggregate
               in the query, and let { key<sub>1</sub>→Ψ<sub>1</sub>, ...,
```
TallTed, line 9559:
```suggestion
              <i>scalarvals</i>, a partial function (possibly with an empty domain) passed from the aggregate
              in the query; and { key<sub>1</sub>→Ψ<sub>1</sub>, ...,
```

spec/index.html
```diff
@@ -9555,9 +9555,9 @@ <h4>Aggregate Algebra</h4>
...
-              key<sub>m</sub>→Ψ<sub>m</sub> } be a set of partial functions from keys to
+              key<sub>m</sub>→Ψ<sub>m</sub> } be a partial function from keys to
               solution sequences as produced by the grouping step.</p>
             <p>Aggregation applies the set function func to the given set and produces a
               single value for each key and group of solutions for that key.</p>
```
TallTed, line 9563:
```suggestion
              key<sub>m</sub>→Ψ<sub>m</sub> }, a partial function from keys to
              solution sequences as produced by the grouping step.</p>
            <p>Aggregation applies the set function `func` to the given set and produces a
              single value for each key and a group of solutions for that key.</p>
```

TallTed (changes requested):
The top comment should be treated on its own; the latter 3 suggestions should be treated as one (they're only split because GitHub doesn't let one suggestion include lines that are shown as deleted).

kasei:
Thanks, @Tpt . Was planning on merging this today :D

Issue #24, opened 2023-03-06
SPARQL Query Errata (5/5)


afs:
https://www.w3.org/2013/sparql-errata

- [ ] [errata-query-20](https://www.w3.org/2013/sparql-errata#errata-query-20)
- [ ] [errata-query-21](https://www.w3.org/2013/sparql-errata#errata-query-21)
- [ ] [errata-query-22](https://www.w3.org/2013/sparql-errata#errata-query-22)
- [ ] [errata-query-22](https://www.w3.org/2013/sparql-errata#errata-query-23)

PR #122, opened 2023-09-21
Fixes Sum example to be on a sequence of lists

Tpt:
The other examples where fixed but Sum seemed to have slipped

Errata https://www.w3.org/2013/sparql-errata#errata-query-22 Issue #24

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/122.html" title="Last updated on Sep 21, 2023, 4:54 PM UTC (0dc4893)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/122/bced028...0dc4893.html" title="Last updated on Sep 21, 2023, 4:54 PM UTC (0dc4893)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -9708,7 +9708,7 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
                 1<br>
                 Sum(L) = op:numeric-add(L<sub>1</sub>, 0) if <a href="#defn_Card">Card</a>(L) = 1<br>
                 Sum(L) = "0"^^xsd:integer if <a href="#defn_Card">Card</a>(L) = 0</p>
-              <p>In this way, Sum( (1, 2, 3) ) = op:numeric-add(1, op:numeric-add(2,
+              <p>In this way, Sum([(1), (2), (3)]) = op:numeric-add(1, op:numeric-add(2,
                 op:numeric-add(3, 0))).</p>
             </div>
           </section>


hartig:
It did not slip when I touched these parts in PR #98. In contrast, earlier the example said:

"In this way, _Sum_(1,2,3) = ..."

and I changed it to:

"In this way, _Sum_( (1,2,3) ) = ..."

Notice that the function name _Sum_ is actually overloaded in this whole definition; that is, there are actually two functions called _Sum_:

One of them is the one that the definition is meant to define, which takes as argument a sequence S that consists of lists of values; i.e., S is of the form [ L1, L2, ..., Ln ] where each Li is a list of values, and [ (1), (2), (3) ] is a concrete example of such a sequence (containing exactly three lists where each of them contains one value). 

The other function called _Sum_ takes as argument a single list of of values; i.e., (1,2,3) is a concrete example of such a list (containing three values). This second _Sum_ is introduced in the definition as a helper construct to define the first _Sum_ function. That is, the line "_Sum_(S) = _Sum_(L)" in the definition defines the first _Sum_ function by using the second one, and the three lines that start with "_Sum_(L) = ..." define the second _Sum_ function.

Now, the way I wrote the example assumes that this is an example of the second _Sum_ function (i.e., the one that takes a single list of values as argument). The change in your PR assumes that this is an example of the first _Sum_ function. In that sense, both ways of writing the example are correct because of the ambiguity of the function name.

A way to resolve this ambiguity would be to introduce a different name for the second _Sum_ function (the helper construct). For instance, it may be renamed to _Sum'_. Perhaps it is useful to do that? Let me know if you want me to do it and I can add a corresponding commit into this PR.

(As a side note, when writing this comment I realized that I did make a mistake in PR #98, but that one is in the definition of the 'Flatten' function. I will create another PR with a fix for this issue.)

TallTed:
Unless I'm misreading things (always possible), there really is only one _Sum()_ function, and it's the argument structure that varies in complexity — it may be a single list, or a list of lists — and the latter form may collapse to the former when the list-of-lists has a single member, i.e., the outer parentheses collapse when they enclose a single inner set.

This is not the simplest thing to explain, but I think it will likely be better understood than _Sum()_ and _Sum'()_ or any other pair of _Sum()_ and _helperSum()_.

hartig:
No, for me it is two different functions. They have different domains. The domain of the first function is the set of lists of lists (or sequences of lists, as it is called in the spec) and the domain of the second function is the set of lists.

Also, the collapsing does not happen magically or implicitly. Instead, that's what the Flatten function is for.

hartig:
I have created PR #124 to illustrate how I think the definition of the Sum function should be improved and how the corresponding example can be presented.

Tpt:
> Notice that the function name Sum is actually overloaded in this whole definition; that is, there are actually two functions called Sum:

Thank you so much for the explanation! I got confused by it. I am closing this MR, #124 is definitely much better.

PR #124, opened 2023-09-22
Improves the section about the Sum set function

hartig:
Related to the discussion in PR #122, this PR here demonstrates how I think the definition of the Sum function should be improved and how the corresponding example can be presented.

Note that this PR uses the feature of ReSpec to explicitly markup symbols in math formulas (as we are planning to use for the whole spec, as per #102). Hence, you can click on every symbol (including function names) and see where else in the section this symbol is used.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/124.html" title="Last updated on Oct 5, 2023, 7:04 PM UTC (ae8c77c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/124/bced028...ae8c77c.html" title="Last updated on Oct 5, 2023, 7:04 PM UTC (ae8c77c)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -9702,13 +9702,21 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
             <div class="defn">
               <p><b>Definition: <span id="defn_aggSum">Sum</span></b></p>
               <pre class="code nohighlight">numeric <var>Sum</var>(sequence <var>S</var>)</pre>
-              <p>|Sum|(|S|) = <var>Sum'</var>(|L|),
-                where |L| = Flatten(|S|)
-                and <var>Sum'</var>(|L|) is defined recursively as follows.</p>
-              <p><var>Sum'</var>(|L|) = op:numeric-add(|L|<sub>1</sub>, <var>Sum'</var>(|L|<sub>2..n</sub>)) if <a href="#defn_Card">Card</a>(|L|) &gt;
-                1<br>
-                <var>Sum'</var>(|L|) = op:numeric-add(|L|<sub>1</sub>, 0) if <a href="#defn_Card">Card</a>(|L|) = 1<br>
-                <var>Sum'</var>(|L|) = "0"^^xsd:integer if <a href="#defn_Card">Card</a>(|L|) = 0</p>
+              <p><var>Sum</var>(<var>S</var>) = <var>Sum'</var>(<var>L</var>),</p>
+              <p>where <var>L</var> = Flatten(<var>S</var>) and
+                <var>Sum'</var>(<var>L</var>) is defined recursively as follows.</p>
+              <ul>
+                <li>If <a href="#defn_Card">Card</a>(<var>L</var>) = 0, then
+                  <var>Sum'</var>(<var>L</var>) = "0"^^<code>xsd:integer</code>.</li>
+                <li>If <a href="#defn_Card">Card</a>(<var>L</var>) = 1, then
+                  <var>Sum'</var>(<var>L</var>) = <a data-cite="XPATH-FUNCTIONS-31#func-numeric-add">op:numeric-add</a>(<var>L</var><sub>1</sub>, 0).</li>
+                <li>If <a href="#defn_Card">Card</a>(<var>L</var>) > 1, then
+                  <var>Sum'</var>(<var>L</var>) = <a data-cite="XPATH-FUNCTIONS-31#func-numeric-add">op:numeric-add</a>(<var>L</var><sub>1</sub>,
+                  <var>Sum'</var>(<var>L</var><sub>2..n</sub>)).</li>
+              </ul>
+              <p>Note that <var>L</var><sub>1</sub> is the first element in
+                <var>L</var>, and <var>L</var><sub>2..n</sub> is <var>L</var>
+                without its first element.</p>
             </div>
             <p>In this way, <var>Sum</var>( [(1), (2), (3)] ) = <var>Sum'</var>( (1, 2, 3) ) =
               op:numeric-add(1, op:numeric-add(2, op:numeric-add(3, 0))).</p>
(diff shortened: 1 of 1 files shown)


hartig:
I just realize that this feature of clicking on symbols in the math formulas does not work in the preview (at least not in my browser). Yet, it works if you load the HTML in the branch of this PR directly into a browser.

spec/index.html
```diff
@@ -9701,16 +9701,17 @@ <h5>Sum</h5>
               be 6.0 (float).</p>
             <div class="defn">
               <p><b>Definition: <span id="defn_aggSum">Sum</span></b></p>
-              <pre class="code nohighlight">numeric Sum(sequence S)</pre>
-              <p>L = Flatten(S)</p>
-              <p>Sum(S) = Sum(L)</p>
-              <p>Sum(L) = op:numeric-add(L<sub>1</sub>, Sum(L<sub>2..n</sub>)) if <a href="#defn_Card">Card</a>(L) &gt;
+              <pre class="code nohighlight">numeric <var>Sum</var>(sequence <var>S</var>)</pre>
+              <p>|Sum|(|S|) = <var>Sum'</var>(|L|),
```
Tpt, line 9705: What about writing: `Sum(S) = Sum'(Flatten(S))`  where Sum'(L) is defined recursively as follows:` instead? I would find it slightly clearer.

Tpt (approved):

spec/index.html
hartig: > What about writing: Sum(S) = Sum'(Flatten(S)) where Sum'(L) is defined recursively as follows:` instead? I would find it slightly clearer.

Not really, because then it would not be clear in the rest of the definition what L is.

hartig:
While the PR in its current form demonstrates my proposal for the 'Sum' set function, there are three other set functions ('Min', 'Max', and 'GroupConcat') whose definition has the same issue as the definition of 'Sum'. I can add corresponding improvements to these three definitions as additional commits in this PR tomorrow.

Nonetheless, if you already have any comments about the proposed way to improve these definitions (as currently demonstrated for 'Sum'), let me know.

hartig:
As mentioned in my previous comment, I now have added the same type of improvements for 'Min', 'Max', and 'GroupConcat' into this PR.

spec/index.html
```diff
@@ -9765,20 +9787,38 @@ <h5>GroupConcat</h5>
...
-              <p>GroupConcat(S, scalarvals) = GroupConcat(L, |separator|)</p>
-              <p>GroupConcat(L, sep) = "" if <a href="#defn_Card">Card</a>(L) = 0</p>
-              <p>GroupConcat(L, sep) = CONCAT("", L<sub>1</sub>) if
-                <a href="#defn_Card">Card</a>(L) = 1</p>
-              <p>GroupConcat(L, sep) = CONCAT(L<sub>1</sub>, sep, GroupConcat(L<sub>2..n</sub>,
-                sep)) if <a href="#defn_Card">Card</a>(L) &gt; 1</p>
-            </div>
-            <p>For example, GroupConcat([("a"), ("b"), ("c")], {"separator" → "."}) = "a.b.c".</p>
+              <pre class="code nohighlight">literal <var>GroupConcat</var>(sequence <var>S</var>, function <var>scalarvals</var>)</pre>
+              <p>If the <var>scalarvals</var> argument is absent from GROUP_CONCAT,
```
TallTed, line 9791:
```suggestion
              <p>If the <var>scalarvals</var> argument is absent from <code>GROUP_CONCAT</code>,
```

spec/index.html
```diff
@@ -9765,20 +9787,38 @@ <h5>GroupConcat</h5>
...
+                <var>GroupConcat'</var>(<var>L</var>, <var>sep</var>),</p>
+              <p>where <var>L</var> = Flatten(<var>S</var>) and
+                <var>GroupConcat'</var>(<var>L</var>, <var>sep</var>)
+                is defined recursively as follows.</p>
+              <ul>
+                <li>If <a href="#defn_Card">Card</a>(<var>L</var>) = 0, then
+                  <var>GroupConcat'</var>(<var>L</var>, <var>sep</var>) = "".</li>
+                <li>If <a href="#defn_Card">Card</a>(<var>L</var>) = 1, then
+                  <var>GroupConcat'</var>(<var>L</var>, <var>sep</var>) =
+                  CONCAT("", <var>L</var><sub>1</sub>).</li>
```
TallTed, line 9810:
```suggestion
                  <code>CONCAT</code>("", <var>L</var><sub>1</sub>).</li>
```

spec/index.html
```diff
@@ -9765,20 +9787,38 @@ <h5>GroupConcat</h5>
...
+                <li>If <a href="#defn_Card">Card</a>(<var>L</var>) > 1, then
+                  <var>GroupConcat'</var>(<var>L</var>, <var>sep</var>) =
+                  CONCAT(<var>L</var><sub>1</sub>, <var>sep</var>, <var>GroupConcat'</var>(<var>L</var><sub>2..n</sub>, <var>sep</var>)).</li>
```
TallTed, line 9813:
```suggestion
                  <code>CONCAT</code>(<var>L</var><sub>1</sub>, <var>sep</var>, <var>GroupConcat'</var>(<var>L</var><sub>2..n</sub>, <var>sep</var>)).</li>
```

spec/index.html
afs: It is good to put the `<code>` on GROUP_CONCAT as per first suggestion where it is in running text.

Here, inside already formatted text for a definition, it's the abstract "concat" like the GroupConcat being defined. The reSpec autolinking should work here. 
I find the `<code>` at this point unhelpful. The style does not, and should not, apply to other names.

spec/index.html
afs: As above.

afs:
Observations rather than change requirement:

A function name with two different signatures (domains) doesn't worry be that much but I'm happy to have it clearly as two functions.

The use of `'` for the secondary function is bit light though I can't think of a naming style that is better - both "SumInner" of "SumCalc".

What ever the style is, it can be easily found later so this PR looks good to me.

afs (approved):

spec/index.html
hartig: > [@afs] Here, inside already formatted text for a definition, it's the abstract "concat" like the GroupConcat being defined. The reSpec autolinking should work here.

Autolinking does not work because there is no `<dfn>CONCAT</dfn> in the document. However, I have now explicitly linked the function name CONCAT in this definition of GroupConcat to the section that defines the CONCAT function. This change is implemented in commit https://github.com/w3c/sparql-query/pull/124/commits/745017a9d1c072ffea2ea39c43581fd54b051a0f

hartig:
@afs 
> A function name with two different signatures (domains) doesn't worry be that much but I'm happy to have it clearly as two functions.

I agree, from the mathematical perspective, there is no problem with overloading the function name. Yet, from a readability perspective such overloading may cause confusion, which is exactly what happened and let @Tpt propose unnecessary changes to the example of Sum in his PR #122. So, that was my motivation for having two separate function names.

> The use of ' for the secondary function is bit light though I can't think of a naming style that is better - both "SumInner" of "SumCalc".

I see. How's about "SumForLists" (and, accordingly, "MinForLists", "MaxForLists", and "GroupConcatForLists") ?

hartig:
Thanks Andy.

What do the other editors think? Better _Sum'_ or _SumList_?

Tpt:
I would tend to prefer `SumList` to avoid the easy to miss `'` but it does not matter much imho.

hartig:
I have now renamed _Sum'_ to _SumList_, _Min'_ to _MinList_, _Max'_ to _MaxList_, and , _GroupConcat'_ to _GCList_ (see commit https://github.com/w3c/sparql-query/pull/124/commits/ae8c77cd9302a67b8bb6863b1ed4c5dce76d8571).

I will wait a few days before I merge this PR, in case the remaining editors who have not looked at this PR (@kasei @rubensworks) want to take a look.

rubensworks (approved):

PR #145, opened 2024-04-04
Update participants, and mark specs as updatable.

gkellogg:
See [resolution](https://www.w3.org/2024/04/04-rdf-star-minutes.html#r02).

For w3c/rdf-star-wg#113.


hartig (approved):

afs (approved):
And all the similar other SPARQL spec updates.

@gkellogg - Thank you!

PR #149, opened 2024-08-12
Migrate from quoted triples to triple terms

rubensworks:
Following the changes in https://github.com/w3c/rdf-concepts/commit/6b7ea76dccfa4ec0d5c9b94cdce7d22dcc0536e5, this PR uses triple terms instead of quoted triples.

For the `TRIPLE` function shorthand, the same syntax was used as in Turtle: `<<( ?s ?p ?o )>>`

This PR does not yet update the grammar, as I believe this was auto-generated before.
@afs I believe you had such a generator? Could you look into running it for these changes? (also for SPARQL Update I guess)
I guess we can do this in a separate PR from this one.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/149.html" title="Last updated on Aug 22, 2024, 6:25 AM UTC (8dcff49)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/149/ca14da9...8dcff49.html" title="Last updated on Aug 22, 2024, 6:25 AM UTC (8dcff49)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -7406,8 +7406,8 @@ WHERE {
         </section>
 
         <!-- QT -->
-        <section id="func-triple-terms">
-          <h4>Functions on Triple Terms</h4>
+        <section id="func-reified-triples">
+          <h4>Functions on Reified Triples</h4>
           <section id="func-triple">
             <h5>TRIPLE</h5>
             <pre class="prototype nohighlight">
@@ -7439,7 +7439,7 @@ class="name">obj</span>)
             <p>
               As a shorthand notation, the <code>TRIPLE</code> function 
               can also be written in the form of a
-              <a href="#rExprQuotedTriple">triple term expression</a>
+              <a href="#rExprTripleTerm">triple term expression</a>
               using <code>&lt;&lt;(</code> and <code>)&gt;&gt;</code>.  There is a
               syntax limitation to this shorthand form: the three elements of
               the triple term expression can only be variables and directly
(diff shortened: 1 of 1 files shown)


afs:
I have a updated grammar in-progress. It includes the current reifier token, multiple reified on stated triples and the multiple annotations work that is now part of Turtle.

[It's written](https://github.com/afs/jena/tree/rdf-star-reifier) - there a few tasks to complete:

* The triples output needs checking - this shows whether the grammar is right or not
* The naming needs to be checked/aligned to the Turtle grammar (QuotedTriple->TripleTerm is done)

hartig (approved):
Looks good!

gkellogg (approved):
Fine with this, but we should probably use "reffing triples" in examples, rather than "triple terms".

TallTed:
> [@gkellogg] Fine with this, but we should probably use "reffing triples" in examples, rather than "triple terms".

I'm sure "reffing triples" is a typo.

I think you meant "reified triples" (which are the objects of triples which have `rdf:reifies` as predicate), but maybe you meant "reifying triples" (which have `rdf:reifies` as predicate).

gkellogg:
> Fine with this, but we should probably use "reifying triples" in examples, rather than "triple terms".

spec/index.html
```diff
@@ -7406,107 +7406,112 @@ <h5>TZ</h5>
         </section>
 
         <!-- QT -->
-        <section id="func-quoted-triples">
-          <h4>Functions on Quoted Triples</h4>
+        <section id="func-reified-triples">
+          <h4>Functions on Reified Triples</h4>
```
hartig, line 7410: @rubensworks  The change that your recent commit https://github.com/w3c/sparql-query/pull/149/commits/118528fb28ee78efe419facac3e903431f0f5b9a implements here is wrong. The functions in this section are not for "reified triples" but for triple terms.
```suggestion
        <section id="func-triple-terms">
          <h4>Functions on Triple Terms</h4>
```

spec/index.html
```diff
@@ -11806,10 +11811,10 @@ <h2>Changes between SPARQL 1.1 Query Language and SPARQL 1.2 Query Language</h2>
             Normative changes:
             <ul>
                 <li>Remove concepts of plain and simple literals, in favor of explicit mentions of xsd:string</li>
-                <li>Update grammar for quoted triples and triple functions in <a href="#sparqlGrammar" class="sectionRef"></a></li>
+                <li>Update grammar for reified triples and triple functions in <a href="#sparqlGrammar" class="sectionRef"></a></li>
```
hartig, line 11814:
```suggestion
                <li>Update grammar for triple terms and triple functions in <a href="#sparqlGrammar" class="sectionRef"></a></li>
```

spec/index.html
```diff
@@ -11806,10 +11811,10 @@ <h2>Changes between SPARQL 1.1 Query Language and SPARQL 1.2 Query Language</h2>
...
                 <li>Migrate XML Schema references to 1.1</li>
                 <li>Update references to XPath from 2.0 to 3.1</li>
-                <li>Add functions on quoted triples to <a href="#func-quoted-triples" class="sectionRef"></a></li>
+                <li>Add functions on reified triples to <a href="#func-reified-triples" class="sectionRef"></a></li>
```
hartig, line 11817:
```suggestion
                <li>Add functions on triple terms to <a href="#func-triple-terms" class="sectionRef"></a></li>
```

hartig (changes requested):

spec/index.html
rubensworks: After @gkellogg's comment above, my reasoning for this change was that readers might be more familiar with the "reified triples" terminology, and would easier find their way to this part in the spec. But you're right that it's not accurate, so let's revert.

hartig (approved):

PR #150, opened 2024-08-16
Update SPARQL grammar for reifier syntax

afs:


spec/index.html
```diff
@@ -10849,622 +10849,692 @@ <h3>Grammar</h3>
...
-                <td><code class="gRuleBody"><span class="token">'&lt;&lt;'</span> <a href="#rDataValueTerm">DataValueTerm</a> ( <a href="#riri">iri</a> | <span class="token">'a'</span> ) <a href="#rDataValueTerm">DataValueTerm</a> <span class="token">'&gt;&gt;'</span></code></td>
+                <td><code class="gRuleBody"><span class="token">'&lt;&lt;'</span> <a href="#rDataValueTerm">DataValueTerm</a> ( <a href="#riri">iri</a> | <span class="token">'a'</span> ) <a href="#rDataValueTerm">DataValueTerm</a> <a href="#rReifierData">ReifierData</a>? <span class="token">'&gt;&gt;'</span></code></td>
               </tr>
 
               <tr style="vertical-align: baseline">
-                <td><code>[111]&nbsp;&nbsp;</code></td>
+                <td><code>[119]&nbsp;&nbsp;</code></td>
+                <td><code><span class="doc-ref" id="rTripleTerm">TripleTerm</span></code></td>
+                <td>&nbsp;&nbsp;::=&nbsp;&nbsp;</td>
+                <td><code class="gRuleBody">L_TRIPLE <a href="#rVarOrTerm">VarOrTerm</a> <a href="#rVerb">Verb</a> <a href="#rVarOrTerm">VarOrTerm</a> R_TRIPLE</code></td>
```
hartig, line 11295: We should make `L_TRIPLE` and `R_TRIPLE` links, as also done for other terminals (e.g., `IRIREF`). That is:
```suggestion
                <td><code class="gRuleBody"><a href="#rL_TRIPLE">L_TRIPLE</a> <a href="#rVarOrTerm">VarOrTerm</a> <a href="#rVerb">Verb</a> <a href="#rVarOrTerm">VarOrTerm</a> <a href="#rR_TRIPLE">R_TRIPLE</a></code></td>
```

hartig, line 11295: We should make `L_TRIPLE` and `R_TRIPLE` links, as also done for other terminals (e.g., `IRIREF`). That is:
```suggestion
                <td><code class="gRuleBody"><a href="#rL_TRIPLE">L_TRIPLE</a> <a href="#rVarOrTerm">VarOrTerm</a> <a href="#rVerb">Verb</a> <a href="#rVarOrTerm">VarOrTerm</a> <a href="#rR_TRIPLE">R_TRIPLE</a></code></td>
```

spec/index.html
```diff
@@ -10849,622 +10849,692 @@ <h3>Grammar</h3>
...
+              </tr>
+
+              <tr style="vertical-align: baseline">
+                <td><code>[120]&nbsp;&nbsp;</code></td>
+                <td><code><span class="doc-ref" id="rTripleTermData">TripleTermData</span></code></td>
+                <td>&nbsp;&nbsp;::=&nbsp;&nbsp;</td>
+                <td><code class="gRuleBody">L_TRIPLE <a href="#rDataValueTerm">DataValueTerm</a> ( <a href="#riri">iri</a> | <span class="token">'a'</span> ) <a href="#rDataValueTerm">DataValueTerm</a> R_TRIPLE</code></td>
```
hartig, line 11302: Same here:
```suggestion
                <td><code class="gRuleBody"><a href="#rL_TRIPLE">L_TRIPLE</a> <a href="#rDataValueTerm">DataValueTerm</a> ( <a href="#riri">iri</a> | <span class="token">'a'</span> ) <a href="#rDataValueTerm">DataValueTerm</a> <a href="#rR_TRIPLE">R_TRIPLE</a></code></td>
```

spec/index.html
```diff
@@ -11476,245 +11546,259 @@ <h3>Grammar</h3>
...
+                <td><code>[179]&nbsp;&nbsp;</code></td>
                 <td><code><span class="doc-ref" id="rANON">ANON</span></code></td>
                 <td>&nbsp;&nbsp;::=&nbsp;&nbsp;</td>
                 <td><code class="gRuleBody">'['  <a href="#rWS">WS</a>* ']'</code></td>
               </tr>
 
               <tr style="vertical-align: baseline">
-                <td><code>[170]&nbsp;&nbsp;</code></td>
+                <td><code>[180]&nbsp;&nbsp;</code></td>
+                <td><code><span class="doc-ref" id="rL_TRIPLE">[&lt;L_TRIPLE</span></code></td>
```
hartig, line 11725:
```suggestion
                <td><code><span class="doc-ref" id="rL_TRIPLE">L_TRIPLE</span></code></td>
```

spec/index.html
```diff
@@ -11476,245 +11546,259 @@ <h3>Grammar</h3>
...
+                <td>&nbsp;&nbsp;::=&nbsp;&nbsp;</td>
+                <td><code class="gRuleBody">'&lt;&lt;('</code></td>
+              </tr>
+
+              <tr style="vertical-align: baseline">
+                <td><code>[181]&nbsp;&nbsp;</code></td>
+                <td><code><span class="doc-ref" id="rR_TRIPLE">[&lt;R_TRIPLE</span></code></td>
```
hartig, line 11732:
```suggestion
                <td><code><span class="doc-ref" id="rR_TRIPLE">R_TRIPLE</span></code></td>
```

spec/index.html
```diff
@@ -10849,622 +10849,692 @@ <h3>Grammar</h3>
...
+                <td><code class="gRuleBody"><a href="#riri">iri</a> | <a href="#rRDFLiteral">RDFLiteral</a> | <a href="#rNumericLiteral">NumericLiteral</a> | <a href="#rBooleanLiteral">BooleanLiteral</a> | <a href="#rVar">Var</a> | <a href="#rExprReifiedTriple">ExprReifiedTriple</a></code></td>
               </tr>
 
               <tr style="vertical-align: baseline">
-                <td><code>[125]&nbsp;&nbsp;</code></td>
-                <td><code><span class="doc-ref" id="rExprQuotedTriple">ExprQuotedTriple</span></code></td>
+                <td><code>[135]&nbsp;&nbsp;</code></td>
+                <td><code><span class="doc-ref" id="rExprReifiedTriple">ExprReifiedTriple</span></code></td>
                 <td>&nbsp;&nbsp;::=&nbsp;&nbsp;</td>
                 <td><code class="gRuleBody"><span class="token">'&lt;&lt;'</span> <a href="#rExprVarOrTerm">ExprVarOrTerm</a> <a href="#rVerb">Verb</a> <a href="#rExprVarOrTerm">ExprVarOrTerm</a> <span class="token">'&gt;&gt;'</span></code></td>
```
hartig, line 11407: Shouldn't that be a pattern for a triple term now? ..rather than a pattern for a reified triple. In other words, should this be:
```suggestion
                <td><code class="gRuleBody"><a href="#rL_TRIPLE">L_TRIPLE</a> <a href="#rExprVarOrTerm">ExprVarOrTerm</a> <a href="#rVerb">Verb</a> <a href="#rExprVarOrTerm">ExprVarOrTerm</a> <a href="#rR_TRIPLE">R_TRIPLE</a></code></td>
```

If you agree, I suggest to change the name of this production as well; `ExpTripleTerm` may be a good option.

hartig (changes requested):

hartig:
Andy, thanks for adapting the grammar! I have added a few comments in my review. Additionally, I have the suspicion that there is an issue in the part of the grammar accepted by the [`ObjectPath`](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/150.html#rObjectPath) production (the link here is directly into the preview of this PR).

For instance, the production accepts the expression
```
<< ?s ?p ?o ~ ?id >> ~ ?id2
```
because it accepts a [`GraphNodePath`](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/150.html#rGraphNodePath) followed by an [`AnnotationPath`](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/150.html#rAnnotationPath), where the `GraphNodePath` in this case is the [`ReifiedTriple`](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/150.html#rReifiedTriple) string (i.e., `<< ?s ?p ?o ~ ?id >>` ) and the `AnnotationPath` is the [`Reifier`](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/150.html#rReifier) string (i.e., `~ ?id2`).

With that expression accepted by the `ObjectPath` production, one may write a triple pattern such as the following.
```
?x ?y  << ?s ?p ?o ~ ?id >> ~ ?id2 .
```
I don't think that makes sense, does it?
(I have to admit that, due to several weeks of vacation, I have not followed the recent discussions that let to the switch from `|` to `~`.)

Other weird-looking things that may be written are
```
?x ?y  << ?s ?p ?o ~ ?id >> ~ ?id2 ~ ?id3 ~ ?id4 .
```
(because [`AnnotationPath`](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/150.html#rAnnotationPath) has the `*`) and
```
?x ?y  << ?s ?p ?o ~ ?id >> ~ ?id2 {| :p : o |} ~ ?id3 .
```

niklasl:
> With that expression accepted by the `ObjectPath` production, one may write a triple pattern such as the following.
> 
> ```
> ?x ?y  << ?s ?p ?o ~ ?id >> ~ ?id2 .
> ```
> 
> I don't think that makes sense, does it? (I have to admit that, due to several weeks of vacation, I have not followed the recent discussions that let to the switch from `|` to `~`.)
> 
> Other weird-looking things that may be written are
> 
> ```
> ?x ?y  << ?s ?p ?o ~ ?id >> ~ ?id2 ~ ?id3 ~ ?id4 .
> ```
> 
> (because [`AnnotationPath`](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/150.html#rAnnotationPath) has the `*`) and
> 
> ```
> ?x ?y  << ?s ?p ?o ~ ?id >> ~ ?id2 {| :p : o |} ~ ?id3 .
> ```

Those do look odd (and probably shouldn't be written like that), but they all (can) make sense. Here's that last one flattened a bit:
```turtle
?x ?y ?id .
<< ?s ?p ?o ~ ?id >> .

<< ?x ?y ?id ~ ?id2 >> .
?id2 :p :o .

<< ?x ?y ?id ~ ?id3 >> .
```

Here's some data with a similar shape, with readable symbol names to provide an informal indication of meaning:
```turtle
<Alice> :says << <Book> :mentions <Bob> ~ <speechact-1> >>
    ~ <recording-2> {| :mentions <Bob> |}
    ~ <recording-3> .
```

Which is short for:
```turtle
<Alice> :says <speechact-1> .
<speechact-1> rdf:reifies <<( <Book> :mentions <Bob> )>> .

<recording-2> rdf:reifies <<( <Alice> :says <speechact-1> )>> .
<recording-2> :mentions <Bob> .

<recording-3> rdf:reifies <<( <Alice> :says <speechact-1> )>> .
```

(Not something I'd put in a primer, but it might perhaps help out here.)

spec/index.html
afs: Thanks checking.

It is a problem with the HTML generator - I'll fix it ASAP so it does not get in the way of other comments.

kasei:
> Is this the agreed-upon version of the shorthand notation now?

And for those of us not joining the semantics TF calls, is there a summary of the new syntax? I've seen it mentioned in passing in emails and on the main WG calls, but don't recall seeing an introduction or summary of the new proposed syntax.

niklasl:
@hartig @kasei I think the discussions originated with https://github.com/w3c/rdf-star-wg/issues/116 (referenced a couple of times on the list). It was worked through more in detail in https://github.com/w3c/rdf-turtle/pull/51.

There's now text and grammar in https://w3c.github.io/rdf-turtle/spec/#reified-triples (thanks to @gkellogg, @TallTed and @afs!).

(It wasn't really part of the semantics discussions IIRC.)

hartig:
Thanks for the pointers Niklas! However:

> There's now text and grammar in https://w3c.github.io/rdf-turtle/spec/#reified-triples (thanks to @gkellogg, @TallTed and @afs!).

The grammar in the Turtle doc is not updated yet, and the text talks only about the new version of the [`reifiedTriple`](https://w3c.github.io/rdf-turtle/spec/#grammar-production-reifiedTriple) production but not about the new version of the annotation syntax that your example uses. Yet, I see that there is an extensive discussion of this new version of the annotation syntax in the comments of PR https://github.com/w3c/rdf-turtle/pull/51. So, I assume now that this PR here implements these changes for SPARQL and a similar PR for Turtle will follow. If that's the case, [my earlier comment](https://github.com/w3c/sparql-query/pull/150#issuecomment-2293384620) about the new [ObjectPath](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/150.html#rObjectPath) production in this PR is obsolete.

gkellogg:
> The grammar in the Turtle doc is not updated yet, and the text talks only about the new version of the [`reifiedTriple`](https://w3c.github.io/rdf-turtle/spec/#grammar-production-reifiedTriple) production but not about the new version of the annotation syntax that your example uses. Yet, I see that there is an extensive discussion of this new version of the annotation syntax in the comments of PR [w3c/rdf-turtle#51](https://github.com/w3c/rdf-turtle/pull/51). So, I assume now that this PR here implements these changes for SPARQL and a similar PR for Turtle will follow. If that's the case, [my earlier comment](https://github.com/w3c/sparql-query/pull/150#issuecomment-2293384620) about the new [ObjectPath](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/150.html#rObjectPath) production in this PR is obsolete.

w3c/rdf-turtle#62 does update the annotation syntax, although there is only minimal narrative discussing this. In the case of the annotation syntax, we have the potential to have multiple reifiers and annotation blocks; this doesn't make sense for the reifiedTriple production as that's expected to generate just a single term when parsing.

afs:
> `?x ?y  << ?s ?p ?o ~ ?id >> ~ ?id2 .`

and (reifier declaration)
```
{ <<:s :p :o >> ~ :r . }
```
is not legal.

Temporarily, the [sparql.org query validator](https://www.sparql.org/query-validator.html) is now running a development version with the latest grammar and it will show the generated triples.

Testing has been "light" so far.

The RDF-star features are not fully propagated though the codebase.  Only parse-to-print is likely to work.

kasei:
> @hartig @kasei I think the discussions originated with [w3c/rdf-star-wg#116](https://github.com/w3c/rdf-star-wg/issues/116) (referenced a couple of times on the list). It was worked through more in detail in [w3c/rdf-turtle#51](https://github.com/w3c/rdf-turtle/pull/51).

Thanks, @niklasl. I had missed the subsequent discussion of details (maybe overlooked because I wouldn't have thought the syntax across formats would be a Turtle PR…?), and so had wrongly assumed it was something that had been agreed to in the semantics TF call. Thanks for the pointers!

hartig:
@afs 
> > `?x ?y  << ?s ?p ?o ~ ?id >> ~ ?id2 .`
>
> [...]
> is not legal.

In this case, my [earlier comment](https://github.com/w3c/sparql-query/pull/150#issuecomment-2293384620) about the [`ObjectPath`](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/150.html#rObjectPath) production is actually relevant. That is, by the current definition of `ObjectPath` in this PR, the string above would be accepted by the [`TriplesBlock`](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/150.html#rTriplesBlock) production.

hartig:
> [@gkellogg] https://github.com/w3c/rdf-turtle/pull/62 does update the annotation syntax, although there is only minimal narrative discussing this.

You are right. I was looking at the current main-branch version of the Turtle spec that Niklas was pointing too.

> [@gkellogg] In the case of the annotation syntax, we have the potential to have multiple reifiers and annotation blocks; this doesn't make sense for the reifiedTriple production as that's expected to generate just a single term when parsing.

I agree regarding the reifiedTriple production, and I can also see now that the annotation syntax in https://github.com/w3c/rdf-turtle/pull/62 accepts multiple reifiers and annotation blocks, exactly as the [ObjectPath](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/150.html#rObjectPath) production for SPARQL in this PR here. Now I wonder, however, why @afs [says above](https://github.com/w3c/sparql-query/pull/150#issuecomment-2295168688) that
```
?x ?y  << ?s ?p ?o ~ ?id >> ~ ?id2 .
```
is illegal. The grammar in this PR accepts it and, similarly, the grammar in https://github.com/w3c/rdf-turtle/pull/62 accepts the following Turtle string.
```
:x :y  << :s :p :o ~ :id >> ~ :id2 .
```

afs:
> Now I wonder, however, why @afs [says above](https://github.com/w3c/sparql-query/pull/150#issuecomment-2295168688) that
> 
> ```
> ?x ?y  << ?s ?p ?o ~ ?id >> ~ ?id2 .
> ```
> 
> is illegal.

The declaration form is illegal. I didn't mean to say the `?x ?y  << ?s ?p ?o ~ ?id >> ~ ?id2 .` was illegal. My bad wording.

"""
Note that the declaration form, `{ <<:s :p :o >> ~ :r . }`, is illegal.
"""
This is to keep it being a single triple.

That means it is suitable for writing out Turtle while streaming.

spec/index.html
afs: Fix applied. The name change is `ExprTripleTerm` ("expr" is used for "expression" in the document elsewhere; "exp" has other meaning like "exponent").

hartig: > Fix applied. The name change is `ExprTripleTerm` ("expr" is used for "expression" in the document elsewhere; "exp" has other meaning like "exponent").

Looks good! I will mark this branch of the conversation as resolved.

afs: Correction made and rule name changed to `ExprTripleTerm` ("expr" issed as an abbreviation of expression elsewhere in the document).

hartig:
> The declaration form is illegal. I didn't mean to say the `?x ?y  << ?s ?p ?o ~ ?id >> ~ ?id2 .` was illegal. My bad wording.

Okay, then we are in agreement regarding `?x ?y  << ?s ?p ?o ~ ?id >> ~ ?id2 .`

As for the declaration form ...

> """
> Note that the declaration form, `{ <<:s :p :o >> ~ :r . }`, is illegal.
> """

I see that the new SPARQL grammar in this PR does not accept this. In contrast, however, the new Turtle grammar in https://github.com/w3c/rdf-turtle/pull/62 does! (without the curly braces, of course). In particular, the `triples` production in line 8 of [spec/turtle.bnf](https://github.com/w3c/rdf-turtle/pull/62/files#diff-1e28931d347a7fce06b1d0f640110a04d25d1c0b6cc9be303f11a87567277370) is currently defined as follows.
```
triples     ::= subject predicateObjectList
              | blankNodePropertyList predicateObjectList?
              | reifiedTriple predicateObjectList?
```
The relevant case here is the third option (`reifiedTriple predicateObjectList?`) -- notice the `?` at the end.

afs:
> ```
> triples     ::= subject predicateObjectList
>               | blankNodePropertyList predicateObjectList?
>               | reifiedTriple predicateObjectList?
> ```
> 
> The relevant case here is the third option (`reifiedTriple predicateObjectList?`) -- notice the `?` at the end.

I think that's OK (I don't have a turtle parser to check).

There must be a `predicateObjectList` to get to the annotation rule which is where `~:e` and annotation `{| |}` are allowed after an object.

cc @gkellogg

hartig:
Yes there must be a `predicateObjectList` to get to the annotation rule, but my comment now was not about the annotation syntax.

Instead, it was about only writing a string that matches the `reifiedTriple` production, as a standalone thing (rather than in the subject or object position of something else). By the current grammar in https://github.com/w3c/rdf-turtle/pull/62, the following is a valid Turtle file.
```
<< :s :p :o ~ :r >> .
```
In contrast, the SPARQL grammar in this PR does _not_ accept this in a WHERE clause, such as the following.
```
SELECT * WHERE {
<< :s :p :o ~ :r >> .   # <-- illegal
}
```

afs:
> SELECT * WHERE {
> << :s :p :o ~ :r >> .   # <-- illegal
> }

That's legal. It does not have a refiier after the `>>`.

A reifier after would apply to the triple:

```
_:b rdf:reifies <<:s :p :o >> .
```
and hence the syntax is two triples:

```
_:b rdf:reifies <<:s :p :o >> .
_:c rdf:reifies << _:b rdf:reifies <<:s :p :o >>  >> .
```


The annotation rule of Turtle includes both reifier for a triple and the `{| .. |}`  -- both are optional.
This excludes a refiier after the declaration, not a reifier in the declaration.

Illegal: reifier after:
```
{ << :s :p :o ~ :r >> ~:e }
{ << :s :p :o >> ~:e }
```
Legal:
```
{ << :s :p :o ~ :r >> }
```
Same for Turtle if I read the BNF right and predicateObjectList is necessary to get to the annotation syntax.

This will be eaiesr when we have parsers and not have to read raw BNF!

hartig:
> > SELECT * WHERE {
> >     << :s :p :o ~ :r >> . # <-- illegal
> > }
>
> That's legal.

Ah, yes, I see now. I overlooked that the `PropertyListPath` in `ReifiedTripleBlockPath` may also be the empty string. Apologies for the noise.

> This will be easier when we have parsers and not have to read raw BNF!

:-D  Indeed!

hartig (approved):

niklasl:
> This will be eaiesr when we have parsers and not have to read raw BNF!

FWIW, I've updated LDTR (mainly a TriG-to-JSON-LD transcriber) to support this reifier syntax. Its [LDTR demo app](https://niklasl.github.io/ldtr/demo/) can be used to play with Turtle-star examples.

(It's made using a PEG-based parser generator, so it _should_ be on par with the ENBF. There are some rough corners though, and both the interface and visualization needs more love. Also, over in the JSON-LD WG we're in the process of updating JSON-LD-star, so it's still quite experimental.)

gkellogg:
I've updated my [distiller](http://rdf.greggkellogg.net/distiller?command=serialize&format=turtle&output_format=ntriples&extractAllScripts&rdfstar) with latest versions of N-Triples, N-Quads, and Turtle parsers.

Note that another difference between the Turtle grammar and SPARQL grammar is that SPARQL allows the annotation delimiters without any actual predicate or object:

```SPARQL
SELECT * WHERE {
  :s :p :o {| |} .
}
```

The Turtle grammar requires there to be a `predicateObjectList`.

```turtle
:s :p :o {| :p1 :o1 |}
```

You can achieve the same affect using a reifier instead of an annotation block:

```turtle
:s :p :o ~
```

afs:
Th reason the SPARQL grammar has `{| |}` (empty body) is to help with machine generated syntax. It is a mild inconvenience to track "zero or non-zero".

I'm happy to go either way - they should align.

afs:
`{| |}` removed.

rubensworks (approved):

spec/index.html
```diff
@@ -10457,6 +10457,12 @@ <h3>Grammar</h3>
           <li>The expression argument of an aggregate function can not contain an aggregate function.</li>
           <li>Only custom aggregate functions use the <code>DISTINCT</code> keyword
             in a <a href="#rFunctionCall">function call</a>.</li>
+          <li>A <a href="#rReifier">reifier</a> or
+            <a href="#rAnnotationBlockPath">annotation syntax</a>
+            is only permitted after a triple when the property position is
+            a simple path (IRI, the keyword <code>a</code> or a variable),
```
TallTed, line 10463: The terminal comma prevents potential reader confusion.
```suggestion
            a simple path (an IRI, the keyword <code>a</code>, or a variable),
```

TallTed (changes requested):
Small, for clarity

PR #152, opened 2024-08-28
Use RDF Dataset definition from RDF Concepts

### LINKED ISSUE #28: RDF 1.1 Alignment - RDF Term and RDF Dataset
`RDF Term` and `RDF dataset` were incorporated into RDF 1.1 (after SPARQL 1.1 was published).

We can switch to the definitions there.

`RDF dataset` added blank nodes for graph named (to accommodate JSON-LD).

rubensworks:
Closes #28

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/152.html" title="Last updated on Sep 19, 2024, 6:33 AM UTC (44d060e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/152/ca14da9...44d060e.html" title="Last updated on Sep 19, 2024, 6:33 AM UTC (44d060e)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -3124,9 +3124,9 @@ WHERE {
         predicate and object. Many RDF data stores hold multiple RDF graphs and record information
         about each graph, allowing an application to make queries that involve information from more
         than one graph.</p>
-      <p>A SPARQL query is executed against an <em>RDF Dataset</em> which represents a collection of
+      <p>A SPARQL query is executed against an <a data-cite="RDF12-CONCEPTS#dfn-rdf-dataset">RDF Dataset</a> [[RDF12-CONCEPTS]] which represents a collection of
         graphs. An RDF Dataset comprises one graph, the default graph, which does not have a name, and
-        zero or more named graphs, where each named graph is identified by an IRI. A SPARQL query can
+        zero or more named graphs, where each named graph is identified by an IRI or a blank node. A SPARQL query can
         match different parts of the query pattern against different graphs as described in section
         <a href="#queryDataset">13.3 Querying the Dataset</a>.</p>
       <p>An RDF Dataset may contain zero named graphs; an RDF Dataset always contains one default
@@ -3138,7 +3138,7 @@ WHERE {
         active graph one of all of the named graphs in the dataset for part of the query.</p>
       <section id="exampleDatasets">
         <h3>Examples of RDF Datasets</h3>
-        <p>The definition of RDF Dataset does not restrict the relationships of named and default
+        <p>The <a data-cite="RDF12-CONCEPTS#dfn-rdf-dataset">definition of RDF Dataset</a> [[RDF12-CONCEPTS]] does not restrict the relationships of named and default
           graphs. Information can be repeated in different graphs; relationships between graphs can be
           exposed. Two useful arrangements are:</p>
         <ul>
@@ -7882,23 +7882,16 @@ WHERE {
         abstract query is then evaluated on an RDF dataset.</p>
       <section id="initDefinitions">
         <h3>Initial Definitions</h3>
-        <section id="sparqlDataset">
-          <h4>RDF Dataset</h4>
-          <div class="defn">
-            <p><b>Definition: <span id="defn_RDFDataset">RDF Dataset</span></b></p>
-            <p>An RDF dataset is a set:<br>
-              { G, (&lt;u<sub>1</sub>&gt;, G<sub>1</sub>), (&lt;u<sub>2</sub>&gt;, G<sub>2</sub>), . .
-              . (&lt;u<sub>n</sub>&gt;, G<sub>n</sub>) }<br>
-              where G and each G<sub>i</sub> are graphs, and each &lt;u<sub>i</sub>&gt; is an IRI. Each
-              &lt;u<sub>i</sub>&gt; is distinct.</p>
-            <p>G is called the default graph. (&lt;u<sub>i</sub>&gt;, G<sub>i</sub>) are called named
-              graphs.</p>
-          </div>
+        <section id="activeGraph">
+          <h4>Active Graph</h4>
           <div class="defn">
             <p><b>Definition: <span id="defn_ActiveGraph">Active Graph</span></b></p>
             <p>The <b>active graph</b> is the graph from the dataset used for basic graph pattern
               matching.</p>
           </div>
+        </section>
+        <section id="rdfDatasetMerge">
+          <h4>RDF Dataset Merge</h4>
           <div class="defn">
             <div id="defn_RDFDatasetMerge">
               <b>Definition: RDF Dataset Merge</b>
@@ -8062,7 +8055,7 @@ WHERE {
             <ul>
               <li>E is a <a href="#sparqlAlgebra">SPARQL algebra</a> expression
               </li>
-              <li>DS is an <a href="#defn_RDFDataset">RDF Dataset</a>
+              <li>DS is an <a data-cite="RDF12-CONCEPTS#dfn-rdf-dataset">RDF Dataset</a> [[RDF12-CONCEPTS]]
               </li>
               <li>QF is a <a href="#QueryForms">query form</a>
               </li>
--- a/spec/index.html
+++ b/spec/index.html
@@ -3149,25 +3149,20 @@ WHERE {
         <div class="exampleGroup">
           <b>Example 1:</b>
           <pre class="data nohighlight">
-# <b>Default graph</b>
 PREFIX dc: &lt;http://purl.org/dc/elements/1.1/&gt;
 
 &lt;http://example.org/bob&gt;    dc:publisher  "Bob" .
 &lt;http://example.org/alice&gt;  dc:publisher  "Alice" .
-          </pre>
-          <pre class="data nohighlight">
-# <b>Named graph: http://example.org/bob</b>
-PREFIX foaf: &lt;http://xmlns.com/foaf/0.1/&gt;
 
-_:a foaf:name "Bob" .
-_:a foaf:mbox &lt;mailto:bob@oldcorp.example.org&gt; .
-          </pre>
-          <pre class="data nohighlight">
-# <b>Named graph: http://example.org/alice</b>
-PREFIX foaf: &lt;http://xmlns.com/foaf/0.1/&gt;
+GRAPH &lt;http://example.org/bob&gt; {
+    _:a foaf:name "Bob" .
+    _:a foaf:mbox &lt;mailto:bob@oldcorp.example.org&gt; .
+}
 
-_:a foaf:name "Alice" .
-_:a foaf:mbox &lt;mailto:alice@work.example.org&gt; .
+GRAPH &lt;http://example.org/alice&gt; {
+    _:a foaf:name "Alice" .
+    _:a foaf:mbox &lt;mailto:alice@work.example.org&gt; .
+}
           </pre>
         </div>
         <p>In this example, the default graph contains the names of the publishers of two named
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -7882,23 +7882,16 @@ <h2>Definition of SPARQL</h2>
         abstract query is then evaluated on an RDF dataset.</p>
       <section id="initDefinitions">
         <h3>Initial Definitions</h3>
```
afs, line 7884: Suggestion for a different way to organise the material.

Replace the definition box for "RDF Dataset" with a paragraph that says the definition comes from RDF concepts. 

As it is important, repeat the definition written in the SPARQL style but not as a definition box. The SPARQL form is used in dataset merge that follows.

Active graph and merge can be in the same section as currently.
This leaves the numbering the same.

afs:
We should align to the RDF Concepts definition - that is to be expected; it was spec timing that caused the divergence.

It is not editorial though because of the blank node change.

The consequences are limited by the fact that allowing blanks in syntax mostly has no consequence because the blank node is not one in the graph. `GRAPH []`/ `GRAPH _:b` could be allowed with that blank node being a variable like it is in BGPs.

An info box in the RDF dataset could mention this.

To work though though the specs:

1. Should `FROM NAMED` allow a blank node?
1. Should <code>GRAPH <i>blanknode</i></code> be allowed?
1. Update: should `INSERT DATA` syntax allow a blank node? It would be a fresh blank node.
1. Update: should `LOAD` allow dataset forms.

The main use case for blank node named graph comes from loading RDF syntax forms that allow it.

Currently, I only see point 4 as necessary.

See https://github.com/w3c/sparql-update/issues/41

spec/index.html
hartig: Like Andy, I would also prefer to keep the subsection called "RDF Dataset".

hartig: > Replace the definition box for "RDF Dataset" with a paragraph that says the definition comes from RDF concepts.
>
> As it is important, repeat the definition written in the SPARQL style but not as a definition box. The SPARQL form is used in dataset merge that follows.

@rubensworks To achieve the latter you can do the following thing. After the brief paragraph that Andy suggests to have first, you can introduce the more formal definition by another paragraph (_not_ in a definition box) that begins with:

For the following definitions, we capture each RDF dataset as a set: ...

hartig:
Where is the notion of an "RDF Dataset Merge" actually used in the spec? I searched the HTML source code for "defn_RDFDatasetMerge" but couldn't find it anywhere else.

rubensworks:
> Where is the notion of an "RDF Dataset Merge" actually used in the spec? I searched the HTML source code for "defn_RDFDatasetMerge" but couldn't find it anywhere else.

You seem to be right. I also can't find any direct usages.
There are multiple mentions of "RDF merge of the graphs" using the "RDF Merge" definition from RDF semantics: https://w3c.github.io/sparql-query/spec/#unnamedGraph https://w3c.github.io/sparql-query/spec/#exampleDatasets
But since "RDF Dataset Merge" is not directly used (anymore?), we may want to consider removing it.

afs:
> Where is the notion of an "RDF Dataset Merge" actually used in the spec? I searched the HTML source code for "defn_RDFDatasetMerge" but couldn't find it anywhere else.

In update, maybe. It, or the graph store equivalent, will be if LOAD takes quads.

Even if it's not, I think we need a definition somewhere because of the blankc shared between graphs. It could go in RDF concepts if rewritten to be "concepts" style.

hartig:
> 1. Should `FROM NAMED` allow a blank node?

I would say no because I am not sure what this would even mean.

One potential interpretation might be to understand such a blank node as something like a variable, as is the case for blank nodes in BGPs. However, this interpretation doesn't make much sense as SPARQL also doesn't allow something like `FROM NAMED ?v`.

Another potential interpretation might be to understand such a blank node as a way to refer to a named graph that is named with that blank node in the underlying RDF dataset of the triple store. Yet, this wouldn't work because the blank node cannot be given directly but must be written via a blank node identifier, e.g.,  
```
FROM NAMED _:b
```
which the SPARQL parser would then turn into a fresh blank node rather than some blank node that is used as a graph name in some dataset.

> 2. Should `GRAPH blanknode` be allowed?

Probably not. It has almost the same problem as `FROM NAMED`. That is, trying to use this blank node as a graph name does not work because the blank node needs to be written via a blank node identifier. The alternative interpretation (using such a blank node as a form of "anonymous", non-projectable variable) has very little value as the blank node identifier via which this blank node is written should be constrained to be different from the blank node identifiers used in any BGP of the query (exactly as is the case for blank node identifiers in different BGPs).

> 3. Update: should `INSERT DATA` syntax allow a blank node? It would be a fresh blank node.

I don't immediately see a problem with that.

Notice, however, that `DELETE DATA` is a different story.

> 4. Update: should `LOAD` allow dataset forms.

I think that's an orthogonal question.

A related question that is not orthogonal is whether the `INTO GRAPH` feature of `LOAD` should allow a blank node. If it was allowed, the result should always be a new named graph with a new blank node as name (because the blank node identifier via which such a blank node at `INTO GRAPH` is written would be parsed into a fresh blank node).

hartig:
> > Where is the notion of an "RDF Dataset Merge" actually used in the spec? I searched the HTML source code for "defn_RDFDatasetMerge" but couldn't find it anywhere else.
>
> In update, maybe. It, or the graph store equivalent, will be if LOAD takes quads.

Update has it's own notion, see [5.2.1 Dataset-UNION](https://www.w3.org/TR/sparql12-update/#def_datasetUnion) Notice that this one is slightly different from "RDF Dataset Merge" in Query. In particular, Dataset-UNION is defined in terms of set union of graphs whereas "RDF Dataset Merge" is defined in terms of the graph merge operation of RDF.

> Even if it's not, I think we need a definition somewhere because of the blankc shared between graphs. It could go in RDF concepts if rewritten to be "concepts" style.

Yes, RDF Concepts seems to be a more suitable place for such a definition.

afs:
> > 1. Should `FROM NAMED` allow a blank node?
> 
> I would say no because I am not sure what this would even mean.

OK - I was getting ahead of myself and imaging an `AS`.
Nothing for `FROM NAMED`.

afs:
> > 7\. Update: should `LOAD` allow dataset forms.
> 
> I think that's an orthogonal question.
> 
> A related question that is not orthogonal is whether the `INTO GRAPH` feature of `LOAD` should allow a blank node. If it was allowed, the result should always be a new named graph with a new blank node as name (because the blank node identifier via which such a blank node at `INTO GRAPH` is written would be parsed into a fresh blank node).

Good point. It has use for loading data, preparing it, then updating the dataset. There is also the order dataset operations `CLEAR`, `CREATE`, `DROP`, `COPY`, `MOVE`, `ADD`, because one request can be several operations that share a blank node scope.

We don't want to pile too much work - we could agree a design on [sparql-update issues](https://github.com/w3c/sparql-update) where it can done, if time, else leave to the ongoing WG after the 2024-2026 charter end.

spec/index.html
gkellogg: Does it make sense for the example input to have separate boxes for the default and named graphs? At the time, TriG didn't exist (at least in a standardized form), so maybe, for Example  1 in [Examples of RDF Datasets](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/152/ca14da9...09ba854.html#exampleDatasets) could be the following:

```turtle
PREFIX dc: <http://purl.org/dc/elements/1.1/>
PREFIX foaf: <http://xmlns.com/foaf/0.1/>

<http://example.org/bob>    dc:publisher  "Bob" .
<http://example.org/alice>  dc:publisher  "Alice" .

GRAPH <http://example.org/bob> {
  _:a foaf:name "Bob" .
  _:a foaf:mbox <mailto:bob@oldcorp.example.org> .
}

GRAPH <http://example.org/alice> {
  _:b foaf:name "Alice" .
  _:b foaf:mbox <mailto:alice@work.example.org> .
}
```

spec/index.html
```diff
@@ -7882,23 +7882,16 @@ <h2>Definition of SPARQL</h2>
...
-          </div>
+        <section id="activeGraph">
+          <h4>Active Graph</h4>
           <div class="defn">
             <p><b>Definition: <span id="defn_ActiveGraph">Active Graph</span></b></p>
             <p>The <b>active graph</b> is the graph from the dataset used for basic graph pattern
               matching.</p>
           </div>
+        </section>
+        <section id="rdfDatasetMerge">
```
gkellogg, line 7893: With graph [merging](https://www.w3.org/TR/rdf12-semantics/#dfn-merge) defined in RDF Semantics, perhaps it would make sense for RDFS to define Dataset Merging, and reference that, instead.

rubensworks: Created a dedicated issue for this: https://github.com/w3c/sparql-query/issues/155

rubensworks:
All comments have been resolved (except for the RDF Dataset Merge discussion which has been deferred to #155).

Note also the [change](https://github.com/w3c/sparql-query/pull/152/commits/77b9f726746e2bd7ba58c29e46716913857e2d81) in the (now informal) RDF dataset definition which also allows for graphs to be identified by blank nodes, instead of just IRIs.

afs (approved):

domel (approved):

spec/index.html
```diff
@@ -7882,16 +7872,23 @@ <h2>Definition of SPARQL</h2>
...
         <h3>Initial Definitions</h3>
-        <section id="activeGraph">
-          <h4>Active Graph</h4>
+        <section id="sparqlDataset">
+          <h4>RDF Dataset</h4>
+          <div>
+            <p>The concept of an <a data-cite="RDF12-CONCEPTS#dfn-rdf-dataset">RDF Dataset</a> is defined in [[RDF12-CONCEPTS]].</p>
+            <p>For the following definitions, we capture each RDF dataset as a set:<br>
+              { G, (&lt;u<sub>1</sub>&gt;, G<sub>1</sub>), (&lt;u<sub>2</sub>&gt;, G<sub>2</sub>), . .
+              . (&lt;u<sub>n</sub>&gt;, G<sub>n</sub>) }<br>
```
TallTed, line 7881: An ellipsis does not include spaces between the dots; it's just three-dots. In sample data like this, such an ellipse can arguably be placed at the start of the second line (as I have), or at the end of the first line, or both (though one or the other would be more common).
```suggestion
              { G, (&lt;u<sub>1</sub>&gt;, G<sub>1</sub>), (&lt;u<sub>2</sub>&gt;, G<sub>2</sub>),
              ... (&lt;u<sub>n</sub>&gt;, G<sub>n</sub>) }<br>
```

gkellogg (approved):

Tpt (approved):

hartig:
Thanks @rubensworks for resolving the comments.

Do we still want to remove the definition of the notion of "RDF Dataset Merge" [as considered above](https://github.com/w3c/sparql-query/pull/152#issuecomment-2315009904)?

rubensworks:
> Do we still want to remove the definition of the notion of "RDF Dataset Merge" https://github.com/w3c/sparql-query/pull/152#issuecomment-2315009904?

I've created an issue for this (#155). Since there are some differing opinion, it may be better to discuss it there. (maybe even to be discussed during one of our weekly meeting)

hartig:
> I've created an issue for this (https://github.com/w3c/sparql-query/issues/155).

Ah, I must have missed that one. Perfect. Then I am fine as well with merging this PR.

hartig (approved):

TallTed (approved):

PR #159, opened 2024-09-29
GH-157, GH-158: Rules for reified triple subject and object parts

### LINKED ISSUE #157: The SPARQL grammar allows `()` in reified triples and triple terms.
In SPARQL, `()` is a parser token `NIL`. 

It allows `()` to be used as a term in its own right, not just in RDF collections `(....)`. e.g. in `VALUES` and expressions. this is in SPARQL 1.1.

The grammar in the current editors working draft allows:

```
<< _:b ex:x () >>
```

```
<<( _:b ex:x () )>>
```

Turtle is different. Parentheses only occur in `collection ::= '(' object* ')'`.

The SPARQL grammar changed to align with Turtle for reified triples and triple terms.

From: https://github.com/w3c/rdf-tests/issues/140#issuecomment-2376596889

### LINKED ISSUE #158: The SPARQL grammar allows expression syntax triple terms to have triple terms as subjects.
Discovered because of https://github.com/w3c/rdf-tests/pull/144.
```
## expr-tripleterm-02.rq
PREFIX : <http://example.com/ns#>

SELECT * {
  ?s ?p ?o .
  BIND(<<( <<(?s ?p ?o)>> :q :z )>> AS ?t)
}
```
This should made invalid syntax.

afs:
This resolves #157
This resolves #158 

* #157
* #158 
* Add a file with a text version: `sparql.bnf`


rubensworks (approved):
It may be useful to also add your script (or some documentation) to generate the HTML from the BNF in case someone else would change it in the future.

hartig (approved):

gkellogg (approved):

spec/sparql.bnf
```diff
@@ -0,0 +1,284 @@
+QueryUnit                 ::= Query 
+Query                     ::= Prologue<br/>( SelectQuery | ConstructQuery | DescribeQuery | AskQuery )<br/>ValuesClause
```
gkellogg, line 2:
```suggestion
Query ::= Prologue
          ( SelectQuery | ConstructQuery | DescribeQuery | AskQuery )<br/>ValuesClause
```

spec/sparql.bnf
```diff
@@ -0,0 +1,284 @@
...
+SolutionModifier          ::= GroupClause? HavingClause? OrderClause? LimitOffsetClauses?
+GroupClause               ::= 'GROUP' 'BY' GroupCondition+
+GroupCondition            ::= BuiltInCall | FunctionCall | '(' Expression ( 'AS' Var )? ')' | Var 
+HavingClause              ::= 'HAVING' HavingCondition+
+HavingCondition           ::= Constraint
+OrderClause               ::= 'ORDER' 'BY' OrderCondition+
+OrderCondition            ::=    ( ( 'ASC' | 'DESC' ) BrackettedExpression )<br/>| ( Constraint | Var ) 
+LimitOffsetClauses        ::= LimitClause OffsetClause? | OffsetClause LimitClause?
+LimitClause               ::= 'LIMIT' <INTEGER>
+OffsetClause              ::= 'OFFSET' <INTEGER>
```
gkellogg, line 27:
```suggestion
LimitClause               ::= 'LIMIT' INTEGER
OffsetClause              ::= 'OFFSET' INTEGER
```

spec/sparql.bnf
```diff
@@ -0,0 +1,284 @@
...
+Path                      ::= PathAlternative
+PathAlternative           ::= PathSequence ( '|' PathSequence )*
+PathSequence              ::= PathEltOrInverse ( '/' PathEltOrInverse )*
+PathElt                   ::= PathPrimary PathMod?
+PathEltOrInverse          ::= PathElt | '^' PathElt
+PathMod                   ::= '?' | '*' | '+'
+PathPrimary               ::= iri | 'a' | '!' PathNegatedPropertySet | '(' Path ')' 
+PathNegatedPropertySet    ::= PathOneInPropertySet | '(' ( PathOneInPropertySet ( '|' PathOneInPropertySet )* )? ')' 
+PathOneInPropertySet      ::= iri | 'a' | '^' ( iri | 'a' ) 
+Integer                   ::= <INTEGER>
```
gkellogg, line 103:
```suggestion
Integer                   ::= INTEGER
```

spec/sparql.bnf
afs: @gkellogg - as you know , the BNF is generated, not hand edited. I'll make these changes to the script that generates the BNF text otherwise they will get lost.

gkellogg:
There are a number of other places where the sparql.bnf either contains `<br/>` or terminals of the form `<INTEGER>`, which aren't part of the EBNF syntax.

afs:
> It may be useful to also add your script (or some documentation) to generate the HTML from the BNF in case someone else would change it in the future.

Sure - actually, it's alwas been available in the Jena grammar area and is part of a Jena release. The RDF 1.2 is in a development area ATM because I don'#t feel it is stable enough to put in a release.

The process isn't BNF to HTML though. The material is in a JavaCC file which itself is processed to exclude local extensions. Having the JavaCC file is helpful because the grammar is quite large and JavaCC is performing checking, both that rules names do actually tie up and that the grammar is LL(1) and it check for LALR(1), no JavaCC parsing features. It runs the W3C tests on the parser so Ruben's contribution on `rdf-tests` is great to see.

JavaCC itself produces a text file. This is munged into HTML with hrefs, links, HTML escapes into a table. The terminals are manually managed to control whether they get inlined or not, and because there are some escaping and quote choices to make.

There is now a script to do the same to produce W3C BNF (one or two things to fix still but getting there).

This may not be the best way to get text BNF. Extracting the text from the HTML works reasonable well (e.g `htmlq` - any other recommendations?) and that would mean just one set of transforms.

Regardless, the primary grammar is the HTML in the spec document.

spec/sparql.bnf
```diff
@@ -0,0 +1,284 @@
...
+AnnotationPath            ::= ( Reifier | AnnotationBlockPath )*
+AnnotationBlockPath       ::= '{|' PropertyListPathNotEmpty '|}'
+Annotation                ::= ( Reifier | AnnotationBlock )*
+AnnotationBlock           ::= '{|' PropertyListNotEmpty '|}'
+GraphNode                 ::= VarOrTerm |	TriplesNode |	ReifiedTriple
+GraphNodePath             ::= VarOrTerm |	TriplesNodePath |	ReifiedTriple
+VarOrTerm                 ::= Var | iri | RDFLiteral | NumericLiteral | BooleanLiteral | BlankNode | NIL | TripleTerm
+ReifiedTriple             ::= '<<' ReifiedTripleSubject Verb ReifiedTripleObject Reifier? '>>'
+ReifiedTripleSubject      ::= Var | iri | RDFLiteral | NumericLiteral | BooleanLiteral | BlankNode | ReifiedTriple
+ReifiedTripleObject       ::= Var | iri | RDFLiteral | NumericLiteral | BooleanLiteral | BlankNode | ReifiedTriple
```
kasei, line 119: Shouldn't `ReifiedTripleObject` allow `TripleTerm`?

afs:
This is ready and will be more readable when in the document.

It is some fixes rather than brand new material.

PR #174, opened 2024-12-15
Update section number referred to in F&O

afs:
What was section 17.1 is now 19.1 after the Functions and Operator version update.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/174.html" title="Last updated on Dec 15, 2024, 11:19 AM UTC (a2695ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/174/e884490...a2695ee.html" title="Last updated on Dec 15, 2024, 11:19 AM UTC (a2695ee)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -7601,7 +7601,7 @@ class="name">obj</span>)
           [[[XPATH-FUNCTIONS-31]]] [[XPATH-FUNCTIONS-31]] in
           section 
           <a data-cite="XPATH-FUNCTIONS-31#casting-from-primitive-to-primitive"
-          >17.1 Casting from primitive types to primitive types</a>.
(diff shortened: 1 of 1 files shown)


hartig (approved):

Tpt (approved):
Shouldn't we just remove the section number to avoid having this problem again?

afs:
> Shouldn't we just remove the section number to avoid having this problem again?

Could do. This PR is minimal change, and given the WG processes, can IMO be just done once editors are aware a change is being made.

We also put in the section title which is also subject to change. I checked - this is the only place where we put in a section number or title.

PR #177, opened 2025-01-08
Defintions for Values Insertion

afs:
A replacement for `substitute`.

The context is taken from[SEP-0007](https://github.com/w3c/sparql-dev/blob/main/SEP/SEP-0007/sep-0007.md).

Currently, all new material is in a single section.


spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
 eval(D(G), Slice(L, start, length)) = Slice(eval(D(G), L), start, length)
             </pre>
           </div>
+
+          <!-- ValuesInsertion -->
+          <section>
+            <h3>Values Insertion and `EXISTS`</h3>
+            <p>The following sections contain draft of material for a revised
+              "`exists`" operation.
+            </p>
```
TallTed, line 10378:
```suggestion
            <p>This section contains draft material for a revised
              "`exists`" operation.
            </p>
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
...
+              Background: <a href="https://github.com/w3c/sparql-dev/blob/main/SEP/SEP-0007/sep-0007.md">SPARQL CG SEP-0007</a>
+            </p>
+
+            <section>
+              <h4>Syntax Restriction</h4>
+              <p><i>Additional [[[#sparqlGrammar]]] note:</i></p>
+                <blockquote>
+                Any variable that is assigned to in the graph pattern of `EXISTS`/`NOT EXISTS` must not be in-scope.
+                This applies to `BIND`, variables introduced by `AS` in a `SELECT` clause, variables in a `VALUES`
+                clause, and variables intrucded by `AS` in `GROUP BY`.
```
TallTed, line 10389: If this is a copy-paste blockquote, then the original text should also be edited and/or flagged as an erratum with these corrections.
```suggestion
                Any variable that is assigned in the graph pattern of `EXISTS`/`NOT EXISTS` must not be in-scope.
                This applies to `BIND`, variables introduced by `AS` in a `SELECT` clause, variables in a `VALUES`
                clause, and variables introduced by `AS` in `GROUP BY`.
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
...
+              </blockquote>
+              <p>Extend the "in-scope" rules to include the variables in-scope from the current row:</p>
```
TallTed, line 10391:
```suggestion
              <p>Extend the "in-scope" rules to include the variables that are in-scope for the current row:</p>
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
...
+              <table style="border-collapse: collapse; border-color: #000000; border-spacing:5px; border-width: 1px">
+                <tbody>
+                  <tr>
+                    <th>Syntax Form</th>
+                    <th>In-scope variables</th>
+                  </tr>
+                  <tr>
+                    <td>`EXISTS` and `NOT EXISTS` filters </td>
+                    <td><code>v</code> is in-scope if it is in-scope for pattern to which the `FILTER` is applied.
```
TallTed, line 10400:
```suggestion
                    <td><code>v</code> is in-scope if it is in-scope for the pattern to which the `FILTER` is applied.
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
...
+                    </td>
+                  </tr>
+                </tbody>
+              </table>
+              <p class="note">
+                This restriction means that <a href="#defn_valuesinsertion">values inserted</a>
+                do not conflict with values assigned to varibales with the pattern.
```
TallTed, line 10407:
```suggestion
                do not conflict with values assigned to variables with the pattern.
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
...
+                <p>
+                Define the Projection Expression Variable Remapping `ProjectMap(P,PV)`
+                </p>
+                <pre>ProjectMap(Project(A, PV)) = Project(A1, PV) 
+                                where A1 is the result of applying F
+                                to every variable mentioned in A.
+                </pre>
+                <p>
+                  The Projection Expression Variable Remapping yields an algrebra expression that
+                  evalautes to the same results as the Project argument. No variable of `ProjectMap(Project(A, PV))`
```
TallTed, line 10442:
```suggestion
                  evaluates to the same results as the Project argument. No variable of `ProjectMap(Project(A, PV))`
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
...
+                  that is not in `PV` is mentioned anywhere else in the algebra expression for the query.
+                </p>
+              </div>
+            
+              <p>This process is applied throughout the graph pattern of <code>EXISTS</code>:</p>
+
+              <div class="defn">
+                <b>Definition: <span id="defn_varrename" name="defn_varrename">Variable Remapping</span></b>
+                <p>
+                  For any algebra expression `X`, define the Variable Remapping `PrjMap(X)`of algebra expression `X`:
```
TallTed, line 10452:
```suggestion
                  For any algebra expression `X`, define the Variable Remapping `PrjMap(X)` of algebra expression `X`:
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
...
+                <p>
+                  For a projection algebra operation #sparqlProjection `Project(A, PV)` acting on algreg express `A` and with set of variables `PV`, define
+                  a partial mapping `F` from
+                  `<a href="#sparqlQueryVariables">V</a>`,
+                  the set of all variables, to `V` where:
+                </p>
+                <pre>F(v) = v1 if v is in PV, where v1 is a fresh variable
+F(v) = v if v is not in PV</pre>
+                <p>
+                Define the Projection Expression Variable Remapping `ProjectMap(P,PV)`
```
TallTed, line 10434:
```suggestion
                Define the Projection Expression Variable Remapping `ProjectMap(P, PV)`
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
...
+            
+              <p>This process is applied throughout the graph pattern of <code>EXISTS</code>:</p>
+
+              <div class="defn">
+                <b>Definition: <span id="defn_varrename" name="defn_varrename">Variable Remapping</span></b>
+                <p>
+                  For any algebra expression `X`, define the Variable Remapping `PrjMap(X)`of algebra expression `X`:
+                </p>
+                  <pre>PrjMap(X) = replace all project operations Project(P, PV) 
+                with ProjectMap(P,PV) for each projection in X.</pre>
```
TallTed, line 10455:
```suggestion
                  with ProjectMap(P, PV) for each projection in X.</pre>
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
...
+              </div>
+              <p>
+                The outcome of `PrjMap` is independent of the order of replacement
+                (e.g. bottom-up or top-down).
```
TallTed, line 10459:
```suggestion
                (e.g., bottom-up or top-down).
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
...
+                Replacements may happen several times, depending on recusive order
```
TallTed, line 10460:
```suggestion
                Replacements may happen several times, depending on recursive order,
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
...
+                but each time a replacement is made, the variable not used anywhere else.
+              </p>
+        
+              <p class="note">
+                A variable inside a project expression that is not in the variables projected
+                is not affected by the values insertion operation because it is renamed apart.
```
TallTed, line 10466:
```suggestion
                is not affected by the value insertion operation because it is renamed apart.
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
 eval(D(G), Slice(L, start, length)) = Slice(eval(D(G), L), start, length)
             </pre>
           </div>
+
+          <!-- ValuesInsertion -->
+          <section>
+            <h3>Values Insertion and `EXISTS`</h3>
```
TallTed, line 10375: `Values Insertion` reads very oddly to me. Can this be changed (throughout) to `Value Insertion`?
```suggestion
          <!-- ValueInsertion -->
          <section>
            <h3>Value Insertion and `EXISTS`</h3>
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
...
+              </p>
+            </section>
+
+            <section>
+              <h4>Values Insertion</h4>
+              <div class="defn">
+                <div>
+                  <b>Definition: <span id="defn_valuesinsertion" name="defn_valuesinsertion">Values Insertion</span></b>
+                  <p>
+                    Define the Values Insertion function `ValuesInsert(X, μ)`
```
TallTed, line 10476: If, as I suggested above, `Values Insertion` is being changed to `Value Insertion`, then `ValuesInsert` should likewise be changed to `ValueInsert`
```suggestion
                    Define the Value Insertion function `ValueInsert(X, μ)`
```

spec/index.html
```diff
@@ -10369,7 +10369,151 @@ <h3>Evaluation Semantics</h3>
...
+              <div class="defn">
+                <b>Definition: <span id="x-defn_evalExists" name="x-defn_evalExists">Evaluation of Exists</span></b>
+                <p>
+                  Let `μ` be the current solution mapping for a filter, and `X` a graph pattern,
+                      define the Evaluation of Exists `exists(X)`
+                </p>
+                <pre>exists(X) = true 
+              if eval( D(G), ValuesInsert(PrjMap(X), μ) )
+                 is a non-empty solution sequence.
+exists(X) = false otherwise</pre>
```
TallTed, line 10509:
```suggestion
Otherwise, exists(X) = false.</pre>
```

TallTed (changes requested):

spec/index.html
afs: It is inserting multiple values.

spec/index.html
afs: This is a DRAFT pull request.

These are process notes. They will be removed.

spec/index.html
afs: The style here is that for function definition, with each case being aligned. It is not running text.

spec/index.html
TallTed: Do you use an "eggs beater" or an "egg beater"?

pfps:
In my view "values insertion" is the correct phrase to use.

spec/index.html
```diff
@@ -10424,7 +10424,205 @@ <h3>Evaluation Semantics</h3>
...
+              <div class="defn">
+                <b>Definition: <span id="defn_projmap" name="defn_projmap">Projection Expression Variable Remapping</span></b>
+                <p>
+                  For a projection algebra operation #sparqlProjection `Project(A, PV)` acting on algreg express `A` and with set of variables `PV`, define
+                  a partial mapping `F` from
+                  `<a href="#sparqlQueryVariables">V</a>`,
+                  the set of all variables, to `V` where:
+                </p>
+                <pre>F(v) = v1 if v is in PV, where v1 is a fresh variable
+F(v) = v if v is not in PV</pre>
```
hartig, line 10495: Considering the second line of this definition, it seems to me that `F` is actually not partial.

If I am wrong with this assumption, can you explain which variables are mapped by `F` and which ones are not.

spec/index.html
```diff
@@ -10424,7 +10424,205 @@ <h3>Evaluation Semantics</h3>
...
+                Define the Projection Expression Variable Remapping `ProjectMap(P, PV)`
+                </p>
+                <pre>ProjectMap(Project(A, PV)) = Project(A1, PV) 
+                                where A1 is the result of applying F
+                                to every variable mentioned in A.
+                </pre>
+                <p>
+                  The Projection Expression Variable Remapping yields an algrebra expression that
+                  evaluates to the same results as the Project argument. No variable of `ProjectMap(Project(A, PV))`
+                  that is not in `PV` is mentioned anywhere else in the algebra expression for the query.
```
hartig, line 10506: I don't think that's true based on the current version of this definition. By the current definition, the projection variables are renamed from `A` to `A1`, but not in `PV`. As a consequence, the evaluation of `A1` will not even produce bindings for the projection variables.

Is it possible that you have a mistake in the earlier part of this definition? In particular, I have the feeling that the first line of the definition of `F` should have been as follows.
```
F(v) = v1 if v is mentioned in A but is not in PV , where v1 is a fresh variable
```

spec/index.html
```diff
@@ -10424,7 +10424,205 @@ <h3>Evaluation Semantics</h3>
...
+              </div>
+              <p>This process is applied throughout the graph pattern of <code>EXISTS</code>:</p>
+              <div class="defn">
+                <b>Definition: <span id="defn_varrename" name="defn_varrename">Variable Remapping</span></b>
+                <p>
+                  For any algebra expression `X`, define the Variable Remapping `PrjMap(X)` 
+                  of algebra expression `X`:
+                </p>
+                  <pre>PrjMap(X) = replace all project operations Project(P, PV) 
+                with ProjectMap(P, PV) for each projection in X.</pre>
```
hartig, line 10517: Is the `ProjectMap(P, PV)` here actually meant to be `ProjectMap( Project(P, PV) )` as introduced in the previous definition? If not, how is `ProjectMap(P, PV)` defined?

spec/index.html
```diff
@@ -10424,7 +10424,205 @@ <h3>Evaluation Semantics</h3>
...
+              <h4>Evaluation of EXISTS</h4>
+              <div class="defn">
+                <b>Definition: <span id="x-defn_evalExists" name="x-defn_evalExists">Evaluation of Exists</span></b>
+                <p>
+                  Let `μ` be the current solution mapping for a filter, and `X` a graph pattern,
+                      define the Evaluation of Exists `exists(X)`
+                </p>
+                <pre>exists(X) = true 
+              if eval( D(G), ValuesInScope(PrjMap(X)), μ)
+                 is a non-empty solution sequence.
```
hartig, line 10617: What is `ValuesInScope( .. )`? Where is this defined?

No matter how it is defined, there is another problem here. The formula uses the [eval](https://www.w3.org/TR/sparql12-query/#defn_eval) function with three arguments: i) `D(G)`, ii) the output of `ValuesInScope( .. )`, whatever that might be, and iii) some solution mapping μ. Yet, [eval](https://www.w3.org/TR/sparql12-query/#defn_eval) is not defined for such arguments; it is defined for two arguments: i) `D(G)` and ii) an algebraic expression.

spec/index.html
```diff
@@ -10424,7 +10424,205 @@ <h3>Evaluation Semantics</h3>
...
+                  <p>
+                    During <a href="#sparqlQuery">translation to the SPARQL algebra</a>
+                  </p>
+                  <pre>
+Replace each occurence of `Y` in X where `Y` is one of
+          <a href="#sparqlTranslateBasicGraphPatterns">Basic Graph Pattern</a>,
+          <a href="#sparqlTranslatePathExpressions">Property Path Expression</a>,
+          <a href="#sparqlTranslateGraphPatterns">`Graph(Var, pattern)`</a>,
+          <a href="#https://www.w3.org/TR/sparql12-query/#sparqlTranslateGraphPatterns">Inline Data</a>
+          with `join(Y, BindingInScope())`.</pre>
```
hartig, line 10568: 1. What is X?
2. Instead of Property Path Expression, do you actually mean Property Path Pattern?

spec/index.html
```diff
@@ -10424,7 +10424,205 @@ <h3>Evaluation Semantics</h3>
...
+                to include a function that evaluates to the current row.
+              </p>
+
+              <div class="defn">
+                <div>
+                  <b>Definition: <span id="defn_correlate" name="defn_correlate">Values In Scope</span></b>
+                  <p>
+                    Define the function `BindingInScope()`. 
+                    Evaluation of `BindingInScope()` results in a table of one row,
+                    being the current binding of the enclosing filter.
```
hartig, line 10551: EXIST can be used not only within FILTER.

kasei: Bind, select expression, order by…

spec/index.html
afs: While this is in algebra space, the syntax analogy is `VALUES`. 

The μ can have several bindings, correlating several variables.

spec/index.html
afs: Added: https://github.com/afs/SPARQL-exists/tree/main/tests/exists-6-places

spec/index.html
TallTed: @afs — Every other instance of `(P, PV)` in this document has the interstitial space. Did you apply or reject this suggested change when you marked it resolved? If you rejected, please explain why?

afs: The change has been made - it says "outdated" above in my view of the PR.

spec/index.html
afs: No longer partial.

spec/index.html
afs: Indeed - the definition of F(v) was inverted.

spec/index.html
afs: Trying `InsertValues `.

This text is in "alternative 2" (runtime algebra modification) and the current direction is alternative 1 (translate time and a new algebra operator to get the correlated solution.

spec/index.html
afs: Should be `BindingInScope` from earlier - which needs to be defined. I've added editors notes here and earlier.

I'm wondering if a 3-argument `eval` is the better way to express the steps here but for now I've simply removed the `ValuesInScope`.

PR #179, opened 2025-01-19
Align descriptions of lang functions. Link to RDF Concepts.

afs:
Reword for function definitions for a more consistent style.

Links to RDF Concept for language tag and base direction in lang-related functions.


spec/index.html
```diff
@@ -6327,9 +6329,9 @@ <h5>hasLANG</h5>
             <pre class="prototype nohighlight"> <span class="return">xsd:string</span>  <span class="operator">hasLANG</span> (<span class="type">RDF term</span> <span class="name">term</span>)
             </pre>
             <p>
-              The function `hasLANG` returns <code>true</code> if the
-              argument is a literal with a language
-              tag. Otherwise, the function returns <code>false</code>.
+              Returns `true` if the RDF term argument is a literal with a
```
hartig, line 6332: It seems somewhat inconsistent to me that the argument of `LANG` and of `LANGDIR` is explicitly a literal whereas the argument of this function (and the next) is an RDF term. Do you remember whether there was a reason for this difference?

spec/index.html
```diff
@@ -7275,11 +7281,15 @@ <h5>CONCAT</h5>
             <h5>langMATCHES</h5>
             <pre class="prototype nohighlight"> <span class="return">xsd:boolean</span>  <span class="operator">langMatches</span> (<span class="type"><span class="type">xsd:string</span></span> <span class="name">language-tag</span>, <span class="type"><span class="type">xsd:string</span></span> <span class="name">language-range</span>)
             </pre>
-            <p>Returns <code>true</code> if <code>language-tag</code> (first argument) matches
-              <code>language-range</code> (second argument) per the basic filtering scheme defined in
+
+            <p>Returns `true` if the <a data-cite="RDF12-CONCEPTS#dfn-language-tag">language tag</a>
+              (first argument) matches <code>language-range</code> (second argument).
+              Otherwise, the function returns `false`.
+              MAtchign is performed according to the the basic filtering scheme defined in
```
hartig, line 7288:
```suggestion
              Matching is performed according to the basic filtering scheme defined in
```

spec/index.html
```diff
@@ -12184,14 +12194,19 @@ <h2>Changes between SPARQL 1.1 Query Language and SPARQL 1.2 Query Language</h2>
...
+                <li>Remove concepts of plain and simple literals, in favor of explicit mentions of `xsd:string`</li>
                 <li>Update grammar for triple terms, reifiers, reified triples, annotation syntax, and triple term functions
                   in <a href="#sparqlGrammar" class="sectionRef"></a></li>
-                <li>Update grammar for literal base direction syntax and functions in <a href="#sparqlGrammar" class="sectionRef"></a></li>
+                <li>Update grammar for literal <a data-cite="RDF12-CONCEPTS#dfn-base-direction">base direction</a> syntax</li>
                 <li>Migrate XML Schema references to 1.1</li>
                 <li>Update references to XPath from 2.0 to 3.1</li>
-                <li>Add functions on triple terms to <a href="#func-triple-terms" class="sectionRef"></a></li>
-                <li>Add functions related to base direction and language tag `LANGDIR`, `hasLANG`, hasLANGDIR, `STRLANGDIR`</li>
+                <li>Add functions on triple terms to
```
hartig, line 12203: To be consistent with the next bullet point.
```suggestion
                <li>Add functions related to <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple terms</a> to
```

spec/index.html
afs: The `is*` functions have a signature of `(RDF term)`. 

These two (`hasLANG`, `hasLANGDIR`) are similar. Earlier, they were `is*` but I think that `has*` reads better for these accessors because it is not "is this term a language?", it is "does this term have a language component?" and the question can be extends to any RDF term.

(Un-)language `""` is sort of special (due in part to XML). We can use that to separate "" and error. 

`LANG(<notALiteral>)` returning `""` does not make sense and should be an error. "" (no language tag : literals _may_ have a language tag) and error (it's not a literal -- wrong question) can be different.

hartig: Thanks for the explanation. Makes sense.

kasei:
Looks good beyond the issues @hartig has already called out.

rubensworks (approved):

PR #190, opened 2025-01-29
SPARQL String. Unicode escapes exclude surrogates.

### LINKED ISSUE #188: Make sure SPARQL uses "RDF Strings"
[RDF String](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-string) makes it clearer about legal UTF-8 (no surrogates).

### LINKED ISSUE #189: Unicode escapes should be restricted to exclude surrogates.
e.g, say they can only be for "Unicode scalar values".

Surrogates are not allowed in UTF-8 but `\u` might be used to sneak them in.

Using Unicode escapes, the query can write in the actual unicode codepoint itself using `\U`.

Lone surrogates are wrong.

See #188. "RDF Strings".

afs:
This closes #188
This closes #189 

* Call "SPARQL Request String", "SPARQL string" (original anchor still works)
  "Request" is too oriented to protocol action
* SPARQL string is a RDF string that conforms to the grammar section
* More reSpec on definitions
* Exclude surrogates from unicode escape sequences (clarity - they are not allowed in RDF string)

Used more reSpec

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/190.html" title="Last updated on Jan 30, 2025, 8:50 AM UTC (4038a3b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/190/16f8a7a...4038a3b.html" title="Last updated on Jan 30, 2025, 8:50 AM UTC (4038a3b)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -347,7 +347,7 @@ span.cancast:hover { background-color: #ffa;
       <h2>Abstract</h2>
       <p>
         RDF is a directed, labeled graph data model for representing information in the
-        Web. This specification defines the syntax and semantics of the SPARQL query language for
+        Web. This specification defines the syntax and semantics of the SPARQL Query Language for
         RDF. SPARQL can be used to express queries across diverse data sources, whether the data is
         stored natively as RDF or viewed as RDF via middleware. SPARQL contains capabilities for
         querying required and optional graph patterns along with their conjunctions and
@@ -374,11 +374,11 @@ span.cancast:hover { background-color: #ffa;
         RDF is a directed, labeled graph data model for representing information in the Web. RDF is
         often used to represent, among other things, personal information, social networks, metadata
         about digital artifacts, as well as to provide a means of integration over disparate sources of
-        information. This specification defines the syntax and semantics of the SPARQL query language
+        information. This specification defines the syntax and semantics of the SPARQL Query Language
         for RDF.
       </p>
       <p>
-        The SPARQL query language for RDF is designed to meet the use cases and
+        The SPARQL Query Language for RDF is designed to meet the use cases and
         requirements identified by the RDF Data Access Working Group in [[RDF-DAWG-UC]],
         the SPARQL 1.1 Working Group in [[SPARQL-FEATURES]], and the RDF-star Working Group.
       </p>
@@ -390,7 +390,7 @@ span.cancast:hover { background-color: #ffa;
         </p>
         <p>
           This section of the document, <a href="#introduction">section 1</a>, introduces the SPARQL
-          query language specification. It presents the organization of this specification document and
+          Query Language specification. It presents the organization of this specification document and
           the conventions used throughout the specification.
         </p>
         <p>
@@ -5364,7 +5364,7 @@ WHERE { ?annot  a:annotates  &lt;http://www.w3.org/TR/rdf-sparql-query/&gt; .
       </section>
       <section id="SparqlOps">
         <h3>Function Definitions</h3>
-        <p>This section defines the operators and functions introduced by the SPARQL Query language.
+        <p>This section defines the operators and functions introduced by the SPARQL query language.
           The examples show the behavior of the operators as invoked by the appropriate grammatical
           constructs.</p>
         <section id="func-forms">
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -10551,15 +10570,19 @@ <h3>Codepoint Escape Sequences</h3>
                 <a href="#HEX">HEX</a> <a href="#HEX">HEX</a>
               </td>
               <td>A Unicode code point in the range U+0 to U+FFFF inclusive corresponding to the
-                encoded hexadecimal value.</td>
+                encoded hexadecimal value, excluding U+D800 to U+DFFF, the
```
gkellogg, line 10573: We don't make this restriction in the Turtle (or related) grammars. It's arguably not necessary, as the value space already restricts RDF/SPARQL strings from including bare surrogates. There should probably be some tests that attempt to create strings using such escape sequences and result in syntax errors.

afs: The input is an RDF string but the output isn't. 

By adding the restriction to unicode escapes, the outcome is an RDF string and nothing more needs to be said. Saying it at the point where bad things happen :tm: is IMO clearer. Could be a note.

Otherwise, somewhere should say that EBNF parsing is on a RDF string again, which is really just moving the unicode escape text about.

Slightly different issue in Turtle because of the different way Unicode escapes are handled which is during parsing. But the text there says the outcome is in the range U+0000 to U+FFFF which include surrogates.

Tpt (approved):

hartig (approved):

spec/index.html
```diff
@@ -10510,30 +10510,49 @@ <h4>Notes</h4>
...
       <section id="queryString">
-        <h3>SPARQL Request String</h3>
+        <h3>SPARQL String</h3>
         <p>
-          A <dfn data-lt="SPARQLRequestString">SPARQL Request String</dfn> is
-          a <a>SPARQL Query String</a> or <a>SPARQL Update String</a> and is a Unicode character string
-          (c.f. section 6.1 String concepts of [[CHARMOD]]) in the language defined by the following
-          grammar.</p>
+          <span id="defn_SPARQLRequestString"></span>
+          A <dfn>SPARQL string</dfn> is an
```
TallTed, line 10516: `SPARQL [String|string]` (and similar, e.g., `SPARQL [Update|update]`) — should be made consistent.

rubensworks (approved):

spec/index.html
afs: Done.

spec/index.html
```diff
@@ -5364,7 +5364,7 @@ <h4>Operator Extensibility</h4>
       </section>
       <section id="SparqlOps">
         <h3>Function Definitions</h3>
-        <p>This section defines the operators and functions introduced by the SPARQL Query language.
+        <p>This section defines the operators and functions introduced by the SPARQL query language.
```
TallTed, line 5367: I do *not* understand the variance in capitalization. `SPARQL Query language` here (line 5367) is being changed to `SPARQL query language`, while earlier in the document (line 393), `SPARQL query language` is being changed to `SPARQL Query Language`. Note that neither line contains (part of) a title; they're both body prose.

There are other variances elsewhere in the document. Why do these vary? What is being communicated by the difference in casing, besides confusion?

spec/index.html
```diff
@@ -10510,30 +10510,49 @@ <h4>Notes</h4>
...
+          range <code>#x0 - #x10FFFF</code>.
+        </p>
       </section>
 
       <section id="codepointEscape">
         <h3>Codepoint Escape Sequences</h3>
-        <p>A SPARQL Query String is processed for codepoint escape sequences before parsing by the
+        <p>
+          A <a>SPARQL string</a> is processed for codepoint escape sequences before parsing by the
           grammar defined in EBNF below. The codepoint escape sequences for a SPARQL query string
```
TallTed, line 10553: What's the difference between `a SPARQL string` and `a SPARQL query string`?

afs: A SPARQL string is any input to this section.
The parser has two entry points - query and update. It's two languages that happen to share grammar rules.

A SPARQL query string is specific to the query entry point `QueryUnit`.
A SPARQL update string is specific to the update entry point `UpdateUnit`.

TallTed: According to these documents, the names of these languages are `SPARQL Query` and `SPARQL Update`, not `SPARQL query` nor `SPARQL update`. These capitalizations should be used throughout.

afs: Your question was "What's the difference between a SPARQL string and a SPARQL query string?" 
I have answered your question.  Please acknowledge that.

The defined terminology "SPARQL query string" is highlighted _as a unit_.
Any other non linked text is, maybe, juxtaposition or just helpful.

You have requested small PRs.
This PR makes focused changes.
The improvement is to remove "request" which is wrong - "request" is a protocol matter.

Scope-creep is time consuming.

We are running out of time after going in circles for _months_.
This WG can achieve "make better" but not "make perfect".

kasei (approved):

spec/index.html
TallTed: Yes, you have clarified that a "SPARQL strsing" is a superclass, with subclasses of "SPARQL Query string" and "SPARQL Update string".

I maintain that the capitalization I have hsed here is correct, and is what should be used throughout.

If you disagree with this, perhaps an issue should be opened for this purely editorial question, which can be answered during CR.

spec/index.html
Tpt: sorry for missing this earlier, this prevents code points that can be encoded as paired surrogates to be encoded as as two consecutive escape sequences (one for the high surrogate and one for the low one). I am not sure we allowed that explicitely before so I am not sure it's a big deal. See w3c/rdf-turtle#84

spec/index.html
TallTed: Picked up in https://github.com/w3c/rdf-star-wg/issues/144

PR #194, opened 2025-02-05
Revise RDFterm-equal and function sameTerm. Rename RDFterm-equal to sameValue.

### LINKED ISSUE #25: Rename 'RDFterm-equal'
As mentioned in [this discussion](https://github.com/w3c/sparql-query/pull/17/files#r1136897998),

> It might be better to rename the function. `RDFterm-equal` puts too much on the "RDFterm" and so reads like 'sameTerm' which it isn't.
> [...]
> Suggestions for alternative names:
> * `RDF-value-equals`
> * `sameValue` (c.f. `sameTerm`)
> * `value-equals`

I agree, and I personally like the second alternative most (`sameValue`).

### LINKED ISSUE #187: Refining RDFTerm-equals
#185 shows that `RDFTerm-equals` needs some attention.
(#25 is the issue for renaming this. Trying it out here.)

In the definition of [RDFTerm-equals](https://www.w3.org/TR/sparql12-query/#func-RDFterm-equal) says:

> The function is defined as follows:
> 
> * Returns TRUE if term1 and term2 are equal RDF terms, as defined below.
> * Produces a type error if term1 and term2 are both literals having the same [datatype IRI](https://www.w3.org/TR/rdf12-concepts/#dfn-datatype-iri); this datatype IRI is not in the set of [recognized datatype IRIs](https://www.w3.org/TR/rdf12-concepts/#dfn-recognized-datatype-iri); and the [lexical forms](https://www.w3.org/TR/rdf12-concepts/#dfn-lexical-form) of the two literals are different from one another.
> * Returns FALSE otherwise.

Bullet 2 applies if the terms have the same dataype IRI. What if there are two diffrerent but related datatype IRIs?

"recognized" is about the data generally, and so could apply elsewhere such a comparision (if it matters), not just this function.
Currently, there is a proposal to remove it from RDF Concepts because it is also [defined by RDF Semantics](https://www.w3.org/TR/rdf12-semantics/#dfn-recognize).

We would be better having a SPARQL-focused definition we can limit to this function (e.g. a system may be able to determine equality in some, but not all, cases).

#### Example 1

```sparql
sameValue("0.13"^^xsd:precisionDecimal, "0.13"^^xsd:decimal) 
```

* Not dispatched by the operator mapping table.
* Different terms.
* Not the same datatype IRI.
 
Returns false.  
But they are the same value - and only `error` allows extension.

(`xsd:precisionDecimal` is not derived from `xsd:decimal` nor the other way round)

#### Example 2

```sparql
sameValue("IV"^^my:romanNumeral, "4"^^xsd:decimal)
```
* Not dispatched ("IV"^^my:romanNumeral is not mentioned in [Operand Data Types](https://www.w3.org/TR/sparql12-query/#operandDataTypes)
* Different terms.
* Different datatypes.

Returns false.

`my:romanNumeral` can't be made derived because the lexical space is not digits.

#### Example 3

```sparql
sameValue("2025-01-01T00:00:00Z"^^xsd:dateTime, "2025-01-01T00:00:00Z"^^xsd:dateTimeStamp) 
```
Returns False.

[`xsd:dateTimeStamp`](https://www.w3.org/TR/xmlschema11-2/#dateTimeStamp) is a derived datatype of `xsd:dateTime` but the spec only covers numeric derived types. 

#### Example 4

Systems involving units (different value space to numbers)

```sparql
sameValue("32"^^x:meters, "3200"^^y:centimeters)
```
Not error so can't be an extensions. This is an erratum.

#### Example 5

```sparql
sameValue("abc"^^xsd:string, "IV"^^my:romanNumeral)
```
All it takes is to know to return FALSE is that `my:romanNumeral` is a number or not a legal value, without 

Not error so can't be an extension. This is an erratum IMO.

#### Example 6

```sparql
sameValue("abc"^^xsd:integer, "123"^^xsd:integer)
```

Probably, this should be FALSE because the operator mapping does apply to `=` and a SPARQL processor must understand `xsd:integer`.
It could be `error` because it can't be 123. c.f.

```sparql
sameValue(1 + "abc"^^xsd:integer, "123"^^xsd:integer)
```
which is an error before `sameValue` is invoked.

#### Proposal

Instead of adding one or more cases, I propose defining RDFTerm-equals more in the style of "by contract"
(not exact wording for now): in order:

1. If one or both arguments are known to be ill-typed, then `error`.
2. If the arguments are the `sameTerm` then return TRUE.
3. If the SPARQL processor can determine the values of both terms and it can determine the values are equal, then return TRUE. 
4. If the SPARQL processor can determine the values of the terms can not be equal, the return FALSE.
5. Otherwise `error`.

The 4th case does not necessarily require values themselves because a processor may know they are different value spaces so can't be value-equal. Case 1 removes know-to-ill-typed.

Because of the order of cases 1 and 2, "sameTerm=true" does not imply "sameValue=true". They could be swapped so that is the case.

I don't think there canbe a perfect in all case solution - we are going from terms to values with incomplete knowledge in some cases.

afs:
This closes #187.
This closes #25.

Changes:
* Rename `RDFterm-equal` as `sameValue`
* Reorder `sameTerm`, `sameValue`
* Reword `sameValue` (see #187)
* Bring definition/description of RDF term equality up-to-date - see related issue w3c/rdf-concepts#154
* Replace examples

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/194.html" title="Last updated on Apr 4, 2025, 2:02 PM UTC (0391ac4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/194/1dabcb9...0391ac4.html" title="Last updated on Apr 4, 2025, 2:02 PM UTC (0391ac4)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -243,8 +243,7 @@ table.casting	{ font-size: x-small; }
 span.cancast:hover { background-color: #ffa;
                      color: black; }
 
-.SPARQLoperator	{ background-color: #FFFFbf; /* yellow */
-          }
+.SPARQLoperator	{ font-weight: 600; }
 
       /* ReSpec */
       dfn { font-style: normal ; }
@@ -4665,6 +4664,7 @@ WHERE { ?annot  a:annotates  &lt;http://www.w3.org/TR/rdf-sparql-query/&gt; .
         <p>SPARQL language extensions may treat additional types as being derived from XML schema
           datatypes.</p>
       </section>
+
       <section id="evaluation">
         <h3>Filter Evaluation</h3>
         <p>SPARQL provides a subset of the functions and operators defined by 
@@ -4826,6 +4826,7 @@ WHERE { ?annot  a:annotates  &lt;http://www.w3.org/TR/rdf-sparql-query/&gt; .
             with a datatype of <code>xsd:boolean</code> and a lexical value of "false".</p>
         </section>
       </section>
+
       <section id="OperatorMapping">
         <h3>Operator Mapping</h3>
         <p>The SPARQL grammar identifies a set of operators 
@@ -4925,7 +4926,7 @@ WHERE { ?annot  a:annotates  &lt;http://www.w3.org/TR/rdf-sparql-query/&gt; .
               <th colspan="5" class="subHeading" scope="col">Logical Connectives</th>
             </tr>
             <tr>
-              <th>
+              <th><span id="logical-or-operator"/>
                 <a href="#rConditionalOrExpression" title="ConditionalOrExpression">A <span class="FAOTtoken">||</span> B</a>
               </th>
               <td>
@@ -4934,7 +4935,7 @@ WHERE { ?annot  a:annotates  &lt;http://www.w3.org/TR/rdf-sparql-query/&gt; .
               <td>
                 xsd:boolean <a href="#ebv-arg">(EBV)</a>
               </td>
-              <td class="sparqlOp">
+              <td class="sparqlOp"><span id="logical-and-operator"/>
                 <a href="#func-logical-or" class="SPARQLoperator">logical-or</a>(A, B)
               </td>
               <td>xsd:boolean</td>
@@ -5275,32 +5276,91 @@ WHERE { ?annot  a:annotates  &lt;http://www.w3.org/TR/rdf-sparql-query/&gt; .
               <th colspan="5" class="subHeading" scope="col">SPARQL Tests</th>
             </tr>
             <tr>
-              <th>
-                <a href="#rRelationalExpression" title="RelationalExpression">A <span class="FAOTtoken">=</span> B</a>
-              </th>
-              <td><span class="type RDFterm">RDF term</span></td>
-              <td><span class="type RDFterm">RDF term</span></td>
-              <td class="xpathOp">
-                <a href="#func-sameValue" class="SPARQLoperator">sameValue</a>(A, B)
+              <th><a href="#rRelationalExpression" title="RelationalExpression">A <span class="FAOTtoken">=</span> B</a></th>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-iri">IRI</a></span></td>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-iri">IRI</a></span></td>
+              <td>
+                <a href="#func-sameTerm">sameTerm</a>(A, B)
               </td>
               <td>xsd:boolean</td>
             </tr>
             <tr>
-              <th>
-                <a href="#rRelationalExpression" title="RelationalExpression">A <span class="FAOTtoken">!=</span> B</a>
-              </th>
+              <th><a href="#rRelationalExpression" title="RelationalExpression">A <span class="FAOTtoken">=</span> B</a></th>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-blank-node">Blank Node</a></span></td>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-blank-node">Blank Node</a></span></td>
+              <td>
+                <a href="#func-sameTerm">sameTerm</a>(A, B)
+              </td>
+              <td>xsd:boolean</td>
+            </tr>
+            <tr>
+              <th><a href="#rRelationalExpression" title="RelationalExpression">A <span class="FAOTtoken">=</span> B</a></th>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-triple-term">Triple Term</a></span></td>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-triple-term">Triple Term</a></span></td>
+              <td>
+                ( A.subject = B.subject ) <a href="#logical-and-operator" class="SPARQLoperator">&&</a><br/>
+                ( A.predicate = B.predicate ) <a href="#logical-and-operator" class="SPARQLoperator">&&</a><br/>
+                ( A.object = B.object )
+              </td>
+              <td>xsd:boolean</td>
+            </tr>
+            <tr>
+              <th><a href="#rRelationalExpression" title="RelationalExpression">A <span class="FAOTtoken">!=</span> B</a></th>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-iri">IRI</a></span></td>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-iri">IRI</a></span></td>
+              <td><a data-cite="XPATH-FUNCTIONS-31#func-not">fn:not</a>(<a href="#func-sameTerm">sameTerm</a>(A, B))</td>
+              <td>xsd:boolean</td>
+            </tr>
+            <tr>
+              <th><a href="#rRelationalExpression" title="RelationalExpression">A <span class="FAOTtoken">!=</span> B</a></th>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-blank-node">Blank Node</a></span></td>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-blank-node">Blank Node</a></span></td>
+              <td>
+                <a data-cite="XPATH-FUNCTIONS-31#func-not">fn:not</a>(<a href="#func-sameTerm">sameTerm</a>(A, B)
+              </td>
+              <td>xsd:boolean</td>
+            </tr>
+            <tr>
+              <th><a href="#rRelationalExpression" title="RelationalExpression">A <span class="FAOTtoken">!=</span> B</a></th>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-triple-term">Triple Term</a></span></td>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-triple-term">Triple Term</a></span></td>
+              <td>
+                ( A.subject != B.subject ) <a href="#logical-or-operator" class="SPARQLoperator">||</a><br/>
+                ( A.predicate != B.predicate ) <a href="#logical-or-operator" class="SPARQLoperator">||</a><br/>
+                ( A.object != B.object )
+              </td>
+              <td>xsd:boolean</td>
+            </tr>
+            <tr>
+              <th><a href="#rRelationalExpression" title="RelationalExpression">A <span class="FAOTtoken">=</span> B</a></th>
               <td><span class="type RDFterm">RDF term</span></td>
               <td><span class="type RDFterm">RDF term</span></td>
               <td class="xpathOp">
-                <a data-cite="XPATH-FUNCTIONS-31#func-not">fn:not</a>(<a href="#func-sameValue" class="SPARQLoperator">sameValue</a>(A, B))
+                <a href="#func-sameValue">sameValue</a>(A, B)
+              </td> 
+              <td>xsd:boolean</td>
+            </tr>
+            <tr>
+              <th><a href="#rRelationalExpression" title="RelationalExpression">A <span class="FAOTtoken">!=</span> B</a></th>
+              <td><span class="type RDFterm">RDF term</span></td>
+              <td><span class="type RDFterm">RDF term</span></td>
+              <td class="xpathOp">
+                <a data-cite="XPATH-FUNCTIONS-31#func-not">fn:not</a>(<a href="#func-sameValue">sameValue</a>(A, B))
               </td>
               <td>xsd:boolean</td>
             </tr>
           </tbody>
         </table>
-        <div id="ebv-arg"></div>xsd:boolean function arguments marked with "(EBV)" are
+        <p>
+        <span id="ebv-arg"></span>xsd:boolean function arguments marked with "(EBV)" are
         coerced to xsd:boolean by evaluating the <a href="#ebv">effective boolean value of that
           argument.</a>
+        </p>
+        <p>
+          Operators <span class="SPARQLoperator">=</span> and <span class="SPARQLoperator">!=</span> applied to 
+          <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple terms</a>
+          apply the operator to each of the components.
+        </p>
         <section id="operatorExtensibility">
           <h4>Operator Extensibility</h4>
           <p>SPARQL language extensions may provide additional associations between operators and
@@ -5314,6 +5374,7 @@ WHERE { ?annot  a:annotates  &lt;http://www.w3.org/TR/rdf-sparql-query/&gt; .
                 BY</code></a> clause.</p>
         </section>
       </section>
+
       <section id="SparqlOps">
         <h3>Function Definitions</h3>
         <p>This section defines the operators and functions introduced by the SPARQL query language.
@@ -5678,7 +5739,7 @@ class="expression">expression, ....</span>)
               defined in [[[RDF12-CONCEPTS]]] [[RDF12-CONCEPTS]]; returns FALSE otherwise.</p>
             <p>
               |term1| and |term2| are the
-              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality">same RDF terms</a>
+              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality">same RDF term</a>
               if one of the following is true:
             </p>
             <ul>
@@ -5756,77 +5817,61 @@ class="expression">expression, ....</span>)
 
             <p>This function cannot be used directly in expressions. The purpose
               of this function is to define the semantics of the "=" operator when applied to
-              two RDF terms that do not fall into any of the other, more concrete cases
+              two RDF terms that do not fall into the concrete cases
               covered in the operator mapping table in Section
               <a href="#OperatorMapping" class="sectionRef"></a>.
             </p>
 
-            <p class="ednote" style="background-color: #EEE">
-              Revise for triple terms.
-              <br/>
-              Consider adding non-literal `=` to operator mapping table.
-            </p>
-
             <p>The result of this function is determined by going through the following steps.</p>
+
             <ol>
               <li>If <code>term1</code> and <code>term2</code> are
                 <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality">equal RDF terms</a>,
-                the return TRUE.
-              </li>
-              <li>If both arguments are literals,
-                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
-                and one or both arguments are known to be ill-typed, 
-                then produce a type error.
-              </li>
-              <li>If <code>term1</code> and <code>term2</code> both 
-                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
-                and the SPARQL processor can determine the values are equal,
                 then return TRUE.
               </li>
-              <li>If <code>term1</code> and <code>term2</code> both 
+              <li>If <code>term1</code> or <code>term2</code> is an
+                <a data-cite="RDF12-CONCEPTS#dfn-iri">IRI</a> or a
+                <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a>
+                then return FALSE.
+              <li>If <code>term1</code> and <code>term2</code> are both
+                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
+                and one or both of these literals has a datatype that is
+                not handled by the SPARQL processor,
+                then produce a type error.
+              </li>
+              <li>If <code>term1</code> and <code>term2</code> are both
+                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
+                and one or both of these literals are known to be
+                <a data-cite="RDF12-CONCEPTS#dfn-ill-typed">ill-typed</a>,
+                then produce a type error.
+              </li>
+              <li>
+                `"NaN"^^xsd:double` and `"NaN"^^xsd:float` are considered to 
+                represent the same value.
+                If <code>term1</code> and <code>term2</code> are
+                both "NaN" for either xsd:double or xsd:float, then 
+                return TRUE.
+              </li>
+              <li>If <code>term1</code> and <code>term2</code> are both 
+                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
+                and the SPARQL processor can determine that their the values are equal,
+                then return TRUE.
+              </li>
+              <li>If <code>term1</code> and <code>term2</code> are both 
                 <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
                 and the SPARQL processor can determine the values can not be equal,
                 then return FALSE.
               </li>
-              <li>Otherwise, return FALSE.
+              <li>
+                If <code>term1</code> and <code>term2</code> are both 
+                <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple terms</a>,
+                apply the function `sameValue` pair-wise to each of the components.
+                Return TRUE if each component pair returns TRUE;
+                produce a type error if any component pair produces an error;
+                otherwise return FALSE.
               </li>
+              <li>Otherwise, return FALSE.</li>
             </ol>
-            <div class="ednote" style="background-color: #EEE">
-              <p>
-                The treatment of `NaN` can not be consistent with both "same term means same value" 
-                and "`fn:numeric-equal(NaN, NaN)` is false". So it is an arbitrary choice with
-                arguments for all three possible choices.
-                For pattern matching, `"NaN"^^xsd:double` must match itself and
-                `"NaN"^^xsd:double` does not pattern match `"NaN"^^xsd:float`.
-              </p>
-              <p>
-                The operator mapping for `=` goes to `fn:numeric-compare` which is false.
-                The `sameValue` result could be made an error (or false) as a special case.
-              </p>
-              <p>
-                `"NaN"^^xsd:double = "NaN"^^xsd:float` is also handled by the
-                operator mapping and is false &ndash; `xsd:float` is promoted to `xsd:double`.
-                A reading of IEEE 754 is that `"NaN"^^xsd:double` and `"NaN"^^xsd:float`
-                are the same term also suggests that `sameValue` is true.
-              </p>
-              <p>
-                <b>Proposal 1</b>: follow "sameTerm means sameValue" and `sameValue("NaN"^^xsd:double, "NaN"^^xsd:double)`
-                and `sameValue("NaN"^^xsd:float, "NaN"^^xsd:float)` are <i>true</i>s.
-                <br/>
-                <b>Proposal 2</b>: 
-                follow "there is one NaN" so `sameValue("NaN"^^xsd:double, "NaN"^^xsd:float)` is <i>true</i>.
-              </p>
-              <p>
-                Ref</i>: <a href="https://www.w3.org/TR/xmlschema11-2/#dt-specialvalue">XSD schema 1.1 special values</a>
-                "are members of the datatype's ·value space·. ". It is not clear whether "special value" literals
-                added to the value spaces of `xsd:double` and `xsd:float` are "the same" literals.
-              <p>
-                <i>Ref</i>: <a href="https://en.wikipedia.org/wiki/IEEE_754#Special_values">(wikipedia) IEEE_754 Special Values</a>
-                <i>IEEE 754 specifies a special value called "Not a Number" (NaN)</i>.
-                suggests they are the same.
-                </p>
-            </div>
-
             <div id="sameValue-ill-typed" class="note">
               <p>A literal is 
                 <a data-cite="RDF12-CONCEPTS#dfn-ill-typed">ill-typed</a>
@@ -5847,31 +5892,19 @@ class="expression">expression, ....</span>)
                 The <a href="#OperatorMapping">Operator Mapping</a> for "`=`"
                 is the function
                 <a data-cite="XPATH-FUNCTIONS-31#func-numeric-equal">`fn:numeric-equal`</a>
-                which returns `false` when comparing arguments involving `NaN`.
-                However, `sameTerm("NaN"^^xsd:double, "NaN"^^xsd:double)` is true;
-                `sameValue` treats "NaN" as the "same value".
-                `sameTerm("NaN"^^xsd:double, "NaN"^^xsd:float)` is `true`.
+                which is defined to return `false` when comparing arguments involving `NaN`.
+                However, `sameTerm("NaN"^^xsd:double, "NaN"^^xsd:double)` is true.
+                The function `sameValue` defines `sameValue("NaN"^^xsd:double, "NaN"^^xsd:double)`
+                to be true because the arguments are the same element of the value space.
               </p>
-            </div>
-            <div id="sameValue-old-name" class="note">
-              <p>This function was previously called `RDFterm-equal`.</p>
-            </div>
-
-            <div class="ednote" style="background-color: #EEE">
               <p>
-                A SPARQL processor may support datatypes beyond those required.
-                Suppose `ex:romanNumeral` is the datatype IRI
-                <a href="https://en.wikipedia.org/wiki/Roman_numerals">roman numerals</a>
-                which are integer numbers. The SPARQL processor may then be able
-                to detemine if two literals have the same value, for example,
-                `sameValue(4, "IV"^^ex:romanNumeral)` can then return `true`
-                because argument bot represent the value 4.
-                For `sameValue("abc"^^xsd:string, "IV"^^ex:romanNumeral)`
-                can return `false` because `xsd:string` and `ex:romanNumeral`
-                have different value spaces so the arguments can not be the same value.
+                `sameValue` treats the values of `"NaN"^^xsd:double` and `"NaN"^^xsd:float` as being
+                the same. `sameValue("NaN"^^xsd:double, "NaN"^^xsd:float)` is `true`.
               </p>
             </div>
-
+            <p id="sameValue-zeros" class="note">
+              For xsd:double and xsd:float, `+0`, `-0` and `0` are same value.
+            </p>
             <p id="func-sameValue-note1" class="note">
               An extended implementation may support additional datatypes for literals. An
               implementation processing a query that tests for equivalence of literals with non-recognized datatypes
@@ -5891,22 +5924,6 @@ class="expression">expression, ....</span>)
                     <th>Results</th>
                   </tr>
                   <tr>
-                    <td><code>sameValue(2, +2)</code></td>
-                    <td>true</td>
-                  </tr>
-                  <tr>
-                    <td><code>sameValue(2.0, 2)</code></td>
-                    <td>true</td>
-                  </tr>
-                  <tr>
-                    <td><code>sameValue(2, "2")</code></td>
-                    <td>false</td>
-                  </tr>
-                  <tr>
-                    <td><code>sameValue(4, "iv"^^my:romanNumeral)</code></td>
-                    <td>error, or true by extension</td>
-                  </tr>
-                  <tr>
                     <td><code>sameValue(1e10, "NaN"^^xsd:double)</code></td>
                     <td>false</td>
                   </tr>
@@ -5918,9 +5935,16 @@ class="expression">expression, ....</span>)
                     <td><code>sameValue("NaN"^^xsd:double, "NaN"^^xsd:float)</code></td>
                     <td>true</td>
                   </tr>
+                  <tr>
+                    <td><code>sameValue( &lt;&lt(:s :p 123)&gt;&gt , &lt;&lt(:s :p 123.0)&gt;&gt )</td>
+                    <td>true</td>
+                  </tr>
                 </tbody>
               </table>
             </div>
+            <p id="sameValue-old-name" class="note">
+              This function was called `RDFterm-equal` up until SPARQL 1.1.
+            </p>
           </section>
 
           <section id="func-isIRI">
--- a/spec/index.html
+++ b/spec/index.html
@@ -5281,7 +5281,7 @@ WHERE { ?annot  a:annotates  &lt;http://www.w3.org/TR/rdf-sparql-query/&gt; .
               <td><span class="type RDFterm">RDF term</span></td>
               <td><span class="type RDFterm">RDF term</span></td>
               <td class="xpathOp">
-                <a href="#func-RDFterm-equal" class="SPARQLoperator">RDFterm-equal</a>(A, B)
+                <a href="#func-sameValue" class="SPARQLoperator">sameValue</a>(A, B)
               </td>
               <td>xsd:boolean</td>
             </tr>
@@ -5292,7 +5292,7 @@ WHERE { ?annot  a:annotates  &lt;http://www.w3.org/TR/rdf-sparql-query/&gt; .
               <td><span class="type RDFterm">RDF term</span></td>
               <td><span class="type RDFterm">RDF term</span></td>
               <td class="xpathOp">
-                <a data-cite="XPATH-FUNCTIONS-31#func-not">fn:not</a>(<a href="#func-RDFterm-equal" class="SPARQLoperator">RDFterm-equal</a>(A, B))
+                <a data-cite="XPATH-FUNCTIONS-31#func-not">fn:not</a>(<a href="#func-sameValue" class="SPARQLoperator">sameValue</a>(A, B))
               </td>
               <td>xsd:boolean</td>
             </tr>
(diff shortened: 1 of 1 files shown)


afs:
At the moment this is unfinished, work in progress.

It has editor's notes in the text, in a lurid colour, for further changes. These can be seen in the preview.

The diff isn't very useful except to show the changes are only in one area of the document.

spec/index.html
```diff
@@ -5669,248 +5670,267 @@ <h5>NOT IN</h5>
...
+              of this function is to define the semantics of the "=" operator when applied to
+              two RDF terms that do not fall into any of the other, more concrete cases
+              covered in the operator mapping table in Section
+              <a href="#OperatorMapping" class="sectionRef"></a>.
+            </p>
+
+            <p class="ednote" style="background-color: #EEE">
+              This will be renamed <code>sameValue</code></p>
+
+            <ol>
```
hartig, line 5779:
```suggestion
            <p>The result of this function is determined by going through the following steps.</p>
            <ol>
```

spec/index.html
```diff
@@ -5669,248 +5670,267 @@ <h5>NOT IN</h5>
...
+              <li>If one or both arguments are known to be ill-typed, 
+                then produce a type error.
+              </li>
+              <li>Returns TRUE if <code>term1</code> and <code>term2</code>
+                are <em>equal</em> RDF terms, as defined below.
+              </li>
```
hartig, line 5785: There is no such definition of "_equal_ RDF terms" below.

spec/index.html
```diff
@@ -5669,248 +5670,267 @@ <h5>NOT IN</h5>
...
+              <li>If the SPARQL processor can determine the values of both terms
+                and it can determine the values are equal, then return TRUE.
+              </li>
+              <li>If the SPARQL processor can determine that the values of the terms
+                can not be equal, the return FALSE.</li>
```
hartig, line 5790: Are you assuming that the two terms are literals? If yes, why is it a valid assumption to be made at this point?

spec/index.html
```diff
@@ -5669,248 +5670,267 @@ <h5>NOT IN</h5>
...
+                the SPARQL processor can determine that the values of the
+                arguments are equal or are not equal. If the SPARQL processor
+                can not be sure, it returns `error`.
+              </p>
+            </div>
+            <div id="sameValue-NaN" class="note">
+              <p>
+                The <a href="#OperatorMapping">Operator Mapping</a> for "`=`""
+                is the function
+                <a data-cite="XPATH-FUNCTIONS-31#func-numeric-equal">fn:numeric-equal`</a>
```
hartig, line 5857:
```suggestion
                <a data-cite="XPATH-FUNCTIONS-31#func-numeric-equal">`fn:numeric-equal`</a>
```

spec/index.html
```diff
@@ -5669,248 +5670,267 @@ <h5>NOT IN</h5>
...
+                    <td>true</td>
+                  </tr>
+                  <tr>
+                    <td><code>sameValue(2.0, 2)</code></td>
+                    <td>true</td>
+                  </tr>
+                  <tr>
+                    <td><code>&nbsp;</code></td>
+                    <td></td>
+                  </tr>
```
hartig, line 5919: Did you add this empty row on purpose?

afs:
There are still 4 "editors notes" for work to be done - one of them is waiting on https://github.com/w3c/rdf-concepts/issues/154.

hartig:
> There are still 4 "editors notes" for work to be done - one of them is waiting on https://github.com/w3c/rdf-concepts/issues/154.

Okay, I will work on a PR for https://github.com/w3c/rdf-concepts/issues/154 right away (as PR https://github.com/w3c/rdf-concepts/pull/158 is merged now).

hartig:
> > There are still 4 "editors notes" for work to be done - one of them is waiting on https://github.com/w3c/rdf-concepts/issues/154.
>
> Okay, I will work on a PR for https://github.com/w3c/rdf-concepts/issues/154 right away (as PR https://github.com/w3c/rdf-concepts/pull/158 is merged now).

The PR for https://github.com/w3c/rdf-concepts/issues/154 is ready; see: https://github.com/w3c/rdf-concepts/pull/161

afs:
Status: ready.

There are three editor's notes - these are discussion points and to be removed before merging.

The PR includes renaming `RDFterm-equal` as `sameValue`.

The PR does not include `sameValue` in the grammar.

Tpt (commented):
Thank you for pushing this and the great NaN discussion

spec/index.html
```diff
@@ -5669,248 +5670,253 @@ <h5>NOT IN</h5>
...
+                follow "there is one NaN" so `sameValue("NaN"^^xsd:double, "NaN"^^xsd:float)` is <i>true</i>.
+              </p>
+              <p>
+                Ref</i>: <a href="https://www.w3.org/TR/xmlschema11-2/#dt-specialvalue">XSD schema 1.1 special values</a>
+                "are members of the datatype's ·value space·. ". It is not clear whether "special value" literals
+                added to the value spaces of `xsd:double` and `xsd:float` are "the same" literals.
+              <p>
+                <i>Ref</i>: <a href="https://en.wikipedia.org/wiki/IEEE_754#Special_values">(wikipedia) IEEE_754 Special Values</a>
+                <i>IEEE 754 specifies a special value called "Not a Number" (NaN)</i>.
+                suggests they are the same.
```
Tpt, line 5820: Per this, I think proposal 2 is definitely better and less surprising than proposal 1

spec/index.html
```diff
@@ -5669,248 +5670,253 @@ <h5>NOT IN</h5>
...
-              same <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>; this datatype IRI is <em>not</em> in the
-              set of <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>; and the
-              <a data-cite="RDF12-CONCEPTS#dfn-lexical-form">lexical forms</a> of the two literals are different from one another.</li>
-                <li>Returns FALSE otherwise.</li>
-            </ul>
+
+          <section id="func-sameTerm">
+            <h5>SameTerm</h5>
+            <pre class="prototype nohighlight"> <span class="return">xsd:boolean</span>  <span class="operator">sameTerm</span> (<span class="type"><span class="type RDFterm">RDF term</span></span> <span class="name">term1</span>, <span class="type"><span class="type RDFterm">RDF term</span></span> <span class="name">term2</span>)</pre>
+            <p>Returns TRUE if <code>term1</code> and <code>term2</code> are the same RDF term as
```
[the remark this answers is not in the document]
afs: Yes, this is a repetition of RDF Concepts (and may need realigning after the merge over there).

I'd like to repeat the definitions, noting the primary definitions in Concepts, though because it is central to `sameTerm`.

spec/index.html
```diff
@@ -5669,248 +5670,267 @@ <h5>NOT IN</h5>
...
+            <p class="ednote" style="background-color: #EEE">
+              Finalize when  
+              <a href="https://github.com/w3c/rdf-concepts/issues/154">RDF Concepts issue 154</a>
+              has been resolved.
+            </p>
+
             <p>
-              <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms if any of the following is
-              true:</p>
+              <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms
```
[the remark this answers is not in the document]
afs: changed "equal" to "the same" - there is also a link to `RDF12-CONCEPTS#dfn-rdf-term-equality` (which may have gone in after your comment).

spec/index.html
```diff
@@ -5669,248 +5670,267 @@ <h5>NOT IN</h5>
...
+                    <td>false</td>
+                  </tr>
+                  <tr>
+                    <td><code>sameTerm("abc"@en, "abc"@EN)</code></td>
+                    <td>true</td>
+                  </tr>
+                  <tr>
+                    <td><code>sameTerm("abc"@en--rtl, "abc"@en)</code></td>
+                    <td>true</td>
+                  </tr>
```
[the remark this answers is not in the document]
afs: Oops - cut/paste/revise error.

spec/index.html
```diff
@@ -5669,248 +5670,267 @@ <h5>NOT IN</h5>
...
+              two RDF terms that do not fall into any of the other, more concrete cases
+              covered in the operator mapping table in Section
+              <a href="#OperatorMapping" class="sectionRef"></a>.
+            </p>
+
+            <p class="ednote" style="background-color: #EEE">
+              This will be renamed <code>sameValue</code></p>
+
+            <ol>
+              <li>If one or both arguments are known to be ill-typed,
```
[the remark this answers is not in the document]
afs: Both done.

spec/index.html
afs: Added "if <code>term1</code> and <code>term2</code> are both literals" ... as per @tpt suggestion.

But this needs revising again (not in PR yet) for triple terms.

Only literal-literal sameValue can be extended but it should apply recursively.

IRIs and blank nodes have a loose (not defined) sense of "value" through same term. 

Maybe the Operator Dispatch table could have "IRI = IRI", "blank node" = "blank node" and "triple term = triple term" added. Then `sameValue` is defined for literals.

spec/index.html
afs: Already gone.

spec/index.html
```diff
@@ -5669,248 +5670,253 @@ <h5>NOT IN</h5>
...
-_:c2  :weight       "100"^^t:kilos .
-_:c2  :displacement  "85"^^t:liters .
+            <ol>
+              <li>If one or both arguments are known to be ill-typed, 
+                then produce a type error.
+              </li>
+              <li>Returns TRUE if <code>term1</code> and <code>term2</code> are
+                <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality">equal RDF terms</a>.
+              </li>
+              <li>If the SPARQL processor can determine the values of both terms
```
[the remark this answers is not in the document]
afs: Done

spec/index.html
```diff
@@ -5669,248 +5670,253 @@ <h5>NOT IN</h5>
...
-_:c3  :label        "Container 3" .
-_:c3  :weight       "85"^^t:kilos .
-_:c3  :displacement  "85"^^t:liters .
-              </pre>
-              <div class="queryGroup">
-                <pre class="query nohighlight">
-PREFIX  :      &lt;http://example.org/WMterms#&gt;
-PREFIX  t:     &lt;http://example.org/types#&gt;
+            <div class="ednote" style="background-color: #EEE" >
+              <p>If the function is to become callable as
```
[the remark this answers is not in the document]
afs: Agreed - not callable (this PR).

spec/index.html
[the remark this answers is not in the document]
afs: > Imho this is weird to state "are the same RDF term" and then define "equal RDF terms".

This is because the text in RDF Concepts changed.

spec/index.html
```diff
@@ -5669,248 +5670,259 @@ <h5>NOT IN</h5>
...
+          <section id="func-sameTerm">
+            <h5>sameTerm</h5>
+            <pre class="prototype nohighlight"> <span class="return">xsd:boolean</span>  <span class="operator">sameTerm</span> (<span class="type"><span class="type RDFterm">RDF term</span></span> <span class="name">term1</span>, <span class="type"><span class="type RDFterm">RDF term</span></span> <span class="name">term2</span>)</pre>
+            <p>Returns TRUE if <code>term1</code> and <code>term2</code> are the same RDF term as
+              defined in [[[RDF12-CONCEPTS]]] [[RDF12-CONCEPTS]]; returns FALSE otherwise.</p>
             <p>
-              <code>term1</code> and <code>term2</code> are <em>equal</em> RDF terms if any of the following is
-              true:</p>
+              |term1| and |term2| are the
+              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality">same RDF terms</a>
```
hartig, line 5681:
```suggestion
              In particular, |term1| and |term2| are the
              <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality">same RDF term</a>
```

spec/index.html
```diff
@@ -5669,248 +5670,259 @@ <h5>NOT IN</h5>
...
-              </pre>
-              <div class="queryGroup">
-                <pre class="query nohighlight">
-PREFIX  :      &lt;http://example.org/WMterms#&gt;
-PREFIX  t:     &lt;http://example.org/types#&gt;
+            <p>The result of this function is determined by going through the following steps.</p>
+            <ol>
+              <li>If <code>term1</code> and <code>term2</code> are
+                <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality">equal RDF terms</a>,
+                the return TRUE.
```
hartig, line 5774:
```suggestion
              <li>If <code>term1</code> and <code>term2</code> are
                <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality">equal RDF terms</a>,
                then return TRUE.
```

spec/index.html
```diff
@@ -5669,248 +5670,259 @@ <h5>NOT IN</h5>
...
+            </p>
 
-_:c2  :label        "Container 2" .
-_:c2  :weight       "100"^^t:kilos .
-_:c2  :displacement  "85"^^t:liters .
+            <p class="ednote" style="background-color: #EEE">
+              Revise for triple terms.
+              <br/>
+              Consider adding non-literal `=` to operator mapping table.
+            </p>
```
hartig, line 5768: While I like the idea of adding non-literal `=` to the operator mapping table, I am afraid that this is not so easily done because the operator mapping table still assumes that there is a function for every case. Then, instead of introducing new functions for the non-literal `=` cases, we may as well continue covering these cases here in `sameValue`.

Now to the question about triple terms, I would say the idea is to recursively use the `=` operator for a pairwise comparison of the three elements of the two compared triple terms and, then, decide the result based on the outcome of these three comparisons. The only complication is how errors should be handled. I think there are two options:
* Option 1: If any of the three comparisons produces an error (which, in principle, can only be the comparison of the two objects), then return an error (i.e., no matter what the result of the other two comparisons is).
* Option 2: If any of the three comparisons produces an error and any other of the three comparisons returns FALSE, then return FALSE.

Looking at the text that we had for this case in the CG report (see https://www.w3.org/2021/12/rdf-star.html#rdf-star-termEqual), I have to say that this text is not so clear to me regarding the error case.

My preference is for Option 2.

hartig (commented):
Here are some more comments after the recent round of changes.

spec/index.html
Tpt: On triple terms: +1 for option 2, it's consistent with what happens for logical operators && and łł

spec/index.html
Tpt: > I don't think @afs meant these two proposals to exclude one another.

Indeed, my bad. Thanks!

> the problem is not actually relevant for the definition of `sameValue`

I am not fully convinced of this because it means that because `xsd:float` has an entry in the operator mapping (`fn:numeric-compare`) then it is allowed to defined that `"NaN"^^xsd:float = "NaN"^^xsd:float` is FALSE whereas a custom `ex:myFloat` won't be able to do that because of the `sameValue` definition that mandates that `sameTerm` implies `sameValue`. This introduces a discrepency between built-in datatypes and the others and prevent to consider `sameValue` definition as working for all terms including `xsd:float` literals.

afs: To add to the fun, we also have `0` and `-0`

xsd:double:
> Equality is identity, except that  0 = −0  (although they are not identical) and  NaN ≠ NaN  (although NaN is of course identical to itself).

spec/index.html
afs: RDF Concepts has caught up.
https://w3c.github.io/rdf-concepts/spec/index.html#dfn-rdf-term-equality

spec/index.html
afs: Section revised.

Also added  "IRI = IRI", "blank node" = "blank node" and "triple term = triple term" to the Operator Dispatch table.

spec/index.html
afs: As per @tpt's point about extension, lets have one "NaN" value and it is the same in xsd:doube and xsd:float value spaces.

`sameValue("NaN"^^xsd:double, "NaN"^^xsd:float)` is true.

spec/index.html
```diff
@@ -5669,248 +5670,253 @@ <h5>NOT IN</h5>
...
+            </div>
+            <div id="sameValue-NaN" class="note">
+              <p>
+                The <a href="#OperatorMapping">Operator Mapping</a> for "`=`""
+                is the function
+                <a data-cite="XPATH-FUNCTIONS-31#func-numeric-equal">fn:numeric-equal`</a>
+                which returns `false` when comparing arguments involving `NaN`.
+                However, `sameTerm("NaN"^^xsd:double, "NaN"^^xsd:double)` is true;
+                `sameValue` treats "NaN" as the "same value".
+                `sameTerm("NaN"^^xsd:double, "NaN"^^xsd:float)` is `true`.
```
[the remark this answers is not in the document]
afs: It is highlight the issue. The definition steps now include "NaN" and "0"/"-0". 

Mentioning "0"/"-0" may be unnecessary - that is, it just works - but I need to check this. The only editors note is for this point.

spec/index.html
afs: Change to "same RDF term".

The "In particular" reads to me as a focus on some specific case, whereas this is reiterating all of the RDF Concepts definition.

spec/index.html
afs: Done.

spec/index.html
```diff
@@ -5669,248 +5670,259 @@ <h5>NOT IN</h5>
...
+            <p>The result of this function is determined by going through the following steps.</p>
+            <ol>
+              <li>If <code>term1</code> and <code>term2</code> are
+                <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality">equal RDF terms</a>,
+                the return TRUE.
+              </li>
+              <li>If both arguments are literals,
+                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
+                and one or both arguments are known to be ill-typed, 
+                then produce a type error.
```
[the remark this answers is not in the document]
afs: Done via local editing as part of a batch of other changes.

spec/index.html
```diff
@@ -5669,248 +5670,259 @@ <h5>NOT IN</h5>
...
+              </li>
+              <li>If <code>term1</code> and <code>term2</code> both 
+                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
+                and the SPARQL processor can determine the values are equal,
```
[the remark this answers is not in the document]
afs: Fixed - "can determine that their values are equal"

spec/index.html
afs: I've added IRI/bnode/triple terms to the operator mapping. The definition for `sameValue` should work. Even if we don't make `sameValue` callable, taking the time to make it as complete as possible seems good as we all have our attention on it. There is a trade-off - it does lead to some repetition.

For triple terms in the operator mapping table, I have written out the expression using operator `&&` and naming the triple term components.

Let me know what you think of the revision.

spec/index.html
```diff
@@ -5669,248 +5670,259 @@ <h5>NOT IN</h5>
...
+                then return TRUE.
+              </li>
+              <li>If <code>term1</code> and <code>term2</code> both 
+                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
```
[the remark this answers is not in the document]
afs: Done

spec/index.html
```diff
@@ -5669,248 +5670,259 @@ <h5>NOT IN</h5>
...
+                <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality">equal RDF terms</a>,
+                the return TRUE.
+              </li>
+              <li>If both arguments are literals,
+                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
+                and one or both arguments are known to be ill-typed, 
+                then produce a type error.
+              </li>
+              <li>If <code>term1</code> and <code>term2</code> both 
+                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
```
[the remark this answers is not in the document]
afs: Done

spec/index.html
afs: Done

afs:
Revisions made.

There is one comment above I haven't addressed yet about extension.

There are three ednotes - one to check "0"/"-0", one to revise/remove the example table in `sameValue`, and one remaining about  the extension point.

I hope I haven't missed anything while marking comments resolved - please unresolve if necessary.

spec/index.html
```diff
@@ -5669,248 +5670,253 @@ <h5>NOT IN</h5>
...
+                <a href="https://en.wikipedia.org/wiki/Roman_numerals">roman numerals</a>
+                which are integer numbers. The SPARQL processor may then be able
+                to detemine if two literals have the same value, for example,
+                `sameValue(4, "IV"^^ex:romanNumeral)` can then return `true`
+                because argument bot represent the value 4.
+                For `sameValue("abc"^^xsd:string, "IV"^^ex:romanNumeral)`
+                can return `false` because `xsd:string` and `ex:romanNumeral`
+                have different value spaces so the arguments can not be the same value.
+              </p>
+            </div>
```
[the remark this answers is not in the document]
afs: > My thinking now is: If the operator mapping for = for all these cases goes to fn:numeric-compare, then the problem is not actually relevant for the definition of sameValue (considering sameValue purely as a means to define the semantics of the = operator for cases that are not covered by other entries in the operator mapping table in Sec.17.3). The only point when it becomes relevant is if we make the function callable directly. Right?

Yes.

And even if callable, the fact that "is the same value" and "op:numeric-equal" are different is just how it is.
"NaN" really is "not a number"!

spec/index.html
```diff
@@ -5669,248 +5670,253 @@ <h5>NOT IN</h5>
...
-                </div>
-              </div>
+            <p>Examples:</p>
+            <div class="result">
+              <table>
+                <tbody>
+                  <tr>
+                    <th>sameValue</th>
+                    <th>Results</th>
+                  </tr>
```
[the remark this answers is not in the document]
afs: Done. 

(I won't disagree with removing the examples table completely.)

spec/index.html
```diff
@@ -5274,32 +5276,91 @@ <h3>Operator Mapping</h3>
...
+                ( A.predicate = B.predicate ) <a href="#logical-and-operator" class="SPARQLoperator">&&</a><br/>
+                ( A.object = B.object )
+              </td>
+              <td>xsd:boolean</td>
+            </tr>
+            <tr>
+              <th><a href="#rRelationalExpression" title="RelationalExpression">A <span class="FAOTtoken">!=</span> B</a></th>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-iri">IRI</a></span></td>
+              <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-iri">IRI</a></span></td>
+              <td><a data-cite="XPATH-FUNCTIONS-31#func-not">fn:not</a>(<a href="#func-sameTerm">fn:sameTerm</a>(A, B))</td>
```
Tpt, line 5311:
```suggestion
              <td><a data-cite="XPATH-FUNCTIONS-31#func-not">fn:not</a>(<a href="#func-sameTerm">sameTerm</a>(A, B))</td>
```

spec/index.html
```diff
@@ -5669,248 +5731,213 @@ <h5>NOT IN</h5>
...
-                <span class="name">term2</span> is a <span class="literal type">literal</span>
-                such that the <a data-cite="RDF12-CONCEPTS#dfn-datatype-iri">datatype IRI</a>
-                of each of these two literals is in the set of
-                <a data-cite="RDF12-CONCEPTS#dfn-recognized-datatype-iri">recognized datatype IRIs</a>
-                and both literals have the same <a data-cite="RDF12-CONCEPTS#dfn-literal-value">literal value</a>.
+                For xsd:double and xsd:float, `+0`, `-0` and `0` are same value.
+                If <code>term1</code> and <code>term2</code> each
+                have datatype IRI xsd:double or xsd:float and
+                a lexical form of one of these zeros, then return TRUE.
+              </li>
```
Tpt, line 5850: Is this step useful? If I am not mistaken `-0` and `+0` are equal according to `op:numeric-equal` <q>Equality is identity, except that  0 = −0  (although they are not identical) and  NaN ≠ NaN  (although NaN is of course identical to itself).</q> [XSD datatypes](https://www.w3.org/TR/xmlschema11-2/#double)

spec/index.html
```diff
@@ -5669,248 +5731,213 @@ <h5>NOT IN</h5>
...
               <li>
-                <span class="name">term1</span> is a <span class="literal type">literal</span> and
-                <span class="name">term2</span> is a <span class="literal type">literal</span> such that
-                these two <span class="literal type">literals</span> are equal as per the notion of <a data-cite="RDF12-CONCEPTS#dfn-literal-term-equality">Literal term equality</a> of
-                [[RDF12-CONCEPTS]].
+                "NaN"^^xsd:double and "NaN"^^xsd:float are considered to 
+                represent the same value.
+                If <code>term1</code> and <code>term2</code> are
+                both "NaN" for either xsd:double or xsd:float then 
+                return TRUE.
```
Tpt, line 5843: This issue with specifying this here is that if we consider that if the "operator mapping table" has precedence to this definition then `"NaN"^^xsd:double = "NaN"^^xsd:float` must be evaluated according to [op:numeric-equal](https://www.w3.org/TR/xpath-functions-31/#func-numeric-equal) and so `"NaN"^^xsd:double = "NaN"^^xsd:float`  returns FALSE, opposite to this definition.

Ammending instead the "operator mapping table" to state we follow [op:numeric-equal](https://www.w3.org/TR/xpath-functions-31/#func-numeric-equal)  except that NaN = NaN might be better. Or we can ignore this case and have it implementation defined.

spec/index.html
```diff
@@ -5669,248 +5731,213 @@ <h5>NOT IN</h5>
...
-                <span class="name">term1</span> and <span class="name">term2</span> are the same
-                <span class="bnode type">blank node</span>.
+                If <code>term1</code> and <code>term2</code> are both 
+                <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple terms</a>,
+                apply the function `sameValue` pair-wise to each of the components.
+                Return TRUE if each component pair returns TRUE, otherise return FALSE.
               </li>
-            </ul>
-            <p id="func-RDFterm-equal-note1" class="note">
+              <li>Otherwise, return FALSE.</li>
```
Tpt, line 5867: If I am not mistaken this means that if a datatype is unsupported, then it we return FALSE. For example if `my:romanNumeral` is not supported by the implementation, then `"iv"^^my:romanNumeral = "4"^^xsd:integer` returns FALSE and not an ERROR because none of the previous step match (values of unsupported datatypes are not ill-typed).

I would remove the second step on ill-typed literals and state here something like

```
              <li>If <code>term1</code> and <code>term2</code> are both
                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
                and one or both of these literals are known to be
                <a data-cite="RDF12-CONCEPTS#dfn-ill-typed">ill-typed</a>,
                or its datatype is not handled by the SPARQL implementation,
                then produce a type error.
```

spec/index.html
```diff
@@ -5669,248 +5731,213 @@ <h5>NOT IN</h5>
...
+            <ol>
+              <li>If <code>term1</code> and <code>term2</code> are
+                <a data-cite="RDF12-CONCEPTS#dfn-rdf-term-equality">equal RDF terms</a>,
+                then return TRUE.
+              </li>
+              <li>If <code>term1</code> and <code>term2</code> are both
+                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
+                and one or both of these literals are known to be
+                <a data-cite="RDF12-CONCEPTS#dfn-ill-typed">ill-typed</a>,
+                then produce a type error.
```
Tpt, line 5836: Having this at this position is a bit sad imho because it means that `"true"^^xsd:boolean = "foo"^^xsd:int` is an error whereas it is known to be false because the datatypes are disjoints.

spec/index.html
```diff
@@ -5669,248 +5731,213 @@ <h5>NOT IN</h5>
...
+              </p>
+            </div>
+            <div id="sameValue-NaN" class="note">
+              <p>
+                The <a href="#OperatorMapping">Operator Mapping</a> for "`=`"
+                is the function
+                <a data-cite="XPATH-FUNCTIONS-31#func-numeric-equal">`fn:numeric-equal`</a>
+                which returns `false` when comparing arguments involving `NaN`.
+                However, `sameTerm("NaN"^^xsd:double, "NaN"^^xsd:double)` is true;
+                `sameValue` treats "NaN"^^xsd:double as the "same value".
```
Tpt, line 5891: `sameValue` definition states that the "operator mapping table" has precedence to this definition hence `"NaN"^^xsd:double = "NaN"^^xsd:float` must be evaluated according to [op:numeric-equal](https://www.w3.org/TR/xpath-functions-31/#func-numeric-equal) and so `"NaN"^^xsd:double = "NaN"^^xsd:float`  must return FALSE. If we follow this, the note should be instead something like "note that for `NaN` there is a conflict between the operator mapping table an this definition and that the operator mapping table has precedence, hence `"NaN"^^xsd:double = "NaN"^^xsd:float` is FALSE".

An other option might be to amend instead the "operator mapping table" to state we follow [op:numeric-equal](https://www.w3.org/TR/xpath-functions-31/#func-numeric-equal)  except that NaN = NaN. Or we can ignore this case and have it implementation defined.

Tpt (commented):
I am not sure to grasp everything, so please ignore my comments if I am wrong

spec/index.html
```diff
@@ -5669,248 +5731,213 @@ <h5>NOT IN</h5>
...
+                and the SPARQL processor can determine the values can not be equal,
+                then return FALSE.
               </li>
               <li>
-                <span class="name">term1</span> and <span class="name">term2</span> are the same
-                <span class="bnode type">blank node</span>.
+                If <code>term1</code> and <code>term2</code> are both 
+                <a data-cite="RDF12-CONCEPTS#dfn-triple-term">triple terms</a>,
+                apply the function `sameValue` pair-wise to each of the components.
+                Return TRUE if each component pair returns TRUE, otherise return FALSE.
```
hartig, line 5865:
```suggestion
                Return TRUE if each component pair returns TRUE, otherwise return FALSE.
```

spec/index.html
afs: Type errors are overrideable so an implementation can make this false through it's understanding of value spaces.

The written expression `"true"^^xsd:boolean = "foo"^^xsd:int` would be a type error anyway because the arguments would be evaluated as expressions before calling the `=` dispatch.

`"foo"^^xsd:int` does make the graph [D-unsatisfiable](https://www.w3.org/TR/rdf12-semantics/#dfn-d-unsatisfiability).

Unless there is a problem introduced by rejecting ill-typed literals before points 5 and 6, I'd suggest keeping it this way. Points 5 and 6 would have to change if this changes.

spec/index.html
afs: There isn't a conflict when writing `=` -- the operator mapping does not lead to `sameValue` in this case.

I read XSD/F&O as having the same "NaN" (point in the value spaces) for xsd;double and xsd:float.
"NaN" is discussed in some places without qualification. That isn't a very strong sense of them being the same value but it seems to be the best. IEEE 754is similar.

e.g. F&O:
> > For xs:float and xs:double values, positive zero and negative zero compare equal. INF equals INF, and -INF equals -INF. NaN does not equal itself.

Are you proposing only `sameValue("NaN"^^xsd:double, "NaN"^^xsd:float)` is false? or that all `sameValue(NaN, NaN)` is false?

Previous comment:

https://github.com/w3c/sparql-query/issues/187#issuecomment-2676987040
> I don't think there is a consistent choice and there are arguments for/against any of true, false or error. I prefer the line of argument that "sameTerm implies sameValue" because of term pattern matching.

I think the `sameTerm` (for pattern matching) consistency matters so `sameValue("NaN"^^xsd:double, "NaN"^^xsd:double)` and `sameValue("NaN"^^xsd:float, "NaN"^^xsd:float)` should be true.

spec/index.html
afs: I think it worth calling out this case for clarity (mild advocacy).

(I think this is in an up-PR comment but the GH interface for long PRs makes finding things difficult.)

spec/index.html
afs: Yes, something about unsupported datatypes should be added.

As it currently is, the final case is the case of one of the terms being an IRI or blank node, or there being a kind mismatch.

Style-wise, I prefer to have guard conditions at the start of the list.

So:
* Add a step for "unsupported datatypes" (my preference is separate from point 2, not combined).
* Add a step for one or both being blank node or IRI.

spec/index.html
afs: An implementation can extend because it is an error. An ill-typed literal means the whole graph is bad anyway.

Not sure they are known to be disjoint - there would need to be an assumption of a closed definition on any datatype definition for that.

spec/index.html
Tpt: Yes!

spec/index.html
Tpt: > Yes, those are the choices and there is no perfect choice here :: https://github.com/w3c/sparql-query/issues/187#issuecomment-2676987040.

Yes, definitely. There is no good answer.

> To be as complete as possible in case it becomes callable. (See earlier discussion.)

In this case shouldn't we align with what is happening in `=` i.e. that `sameValue(NaN, NaN)` is false? If not it means `sameValue` and `=` do not have the same behaviors and we end up wih three different equalities in SPARQL (`=`, `sameValue` and `sameTerm`). Imho having`sameValue` and `=` get the exact same output is more important than `sameTerm` implying `sameValue`.

afs: It is for graph matching - an identity relationship. Graph matching is not using an external spec to define "same".

```
?s :p "NaN"^^xsd:double` .
```

Suppose the data is canonicalized on input. Then matching is like `sameValue`.

`?s :p ?X. FILTER (sameValue(?X, "NaN"^^xsd:double)` if callable. Same optimizations of `sameTerm` becomes pattern patch apply.

Why is it important to align `sameValue` and `fn:numeric-equal`?
`=` is always available in expressions.

Tpt: Thank you! Indeed this equivalence is useful and it's sad we don't have it for `=`.

> Why is it important to align `sameValue` and `fn:numeric-equal`?

To have `sameValue(x, y)` as an alias to `x = y` and avoid introducing a third flavor of equality

spec/index.html
Tpt: Make sense! Thank you!

Tpt (approved):
Thank you for pushing this and for answering all my nitpicks.

I am still not convinced that the special case for NaN is worth it while sameValue is kept as only a way to define `=` but it's not a big deal (my bold assumption is that we will never expose sameValue directly).

afs:
Bold.

😄

spec/index.html
```diff
@@ -5669,248 +5731,221 @@ <h5>NOT IN</h5>
...
+            <span id="func-RDFterm-equal"><!-- obsolete id --></span>
+            <h5>sameValue</h5>
+            <pre class="prototype nohighlight">
+              <span class="return">xsd:boolean</span> <span class="operator" style="text-transform: none;">sameValue</span> (<span class="type"><span class="type RDFterm">RDF term</span></span> <span class="name">term1</span>, <span class="type"><span class="type RDFterm">RDF term</span></span> <span class="name">term2</span>)
+            </pre>
+
+            <p>This function cannot be used directly in expressions. The purpose
+              of this function is to define the semantics of the "=" operator when applied to
+              two RDF terms that do not fall into any of the other, more concrete cases
+              covered in the operator mapping table in Section
```
rubensworks, line 5821: Formulation is a bit strange here. As it could be interpreted as there being a category of "other" cases within the operator table.
```suggestion
              two RDF terms that do not fall into the concrete cases
              covered in the operator mapping table in Section
```

spec/index.html
```diff
@@ -5669,248 +5731,221 @@ <h5>NOT IN</h5>
...
+              </p>
+            </div>
+            <div id="sameValue-NaN" class="note">
+              <p>
+                The <a href="#OperatorMapping">Operator Mapping</a> for "`=`"
+                is the function
+                <a data-cite="XPATH-FUNCTIONS-31#func-numeric-equal">`fn:numeric-equal`</a>
+                which returns `false` when comparing arguments involving `NaN`.
+                However, `sameTerm("NaN"^^xsd:double, "NaN"^^xsd:double)` is true;
+                `sameValue` treats "NaN"^^xsd:double as the "same value".
```
rubensworks, line 5897: Some minor suggestions to the phrasing:
```suggestion
                The <a href="#OperatorMapping">Operator Mapping</a> for "`=`"
                uses the function
                <a data-cite="XPATH-FUNCTIONS-31#func-numeric-equal">`fn:numeric-equal`</a>
                which is defined to return `false` when comparing arguments involving `NaN`.
                However, since equality of terms is given precedence over `fn:numeric-equal`,
                and `sameTerm("NaN"^^xsd:double, "NaN"^^xsd:double)` is true;
                `sameValue` treats "NaN"^^xsd:double as the "same value".
```

spec/index.html
```diff
@@ -5669,248 +5731,221 @@ <h5>NOT IN</h5>
...
+                  </tr>
+                  <tr>
+                    <td><code>sameValue( &lt;&lt(:s :p 123)&gt;&gt , &lt;&lt(:s :p 123.0)&gt;&gt )</td>
+                    <td>true</td>
+                  </tr>
+                </tbody>
+              </table>
             </div>
+            <p id="sameValue-old-name" class="note">
+              This function was previously called `RDFterm-equal`.
```
rubensworks, line 5945:
```suggestion
              This function was called `RDFterm-equal` up until SPARQL 1.1.
```

rubensworks (approved):
Just some minor suggestions. Looks good to me in general!

spec/index.html
TallTed: The new line 5898 is not a complete sentence, which it should be. 

There should be a trailing `as xyz` that says what it is that ``` `"NaN"^^xsd:double``` is `treated as the "same value"` *of*.

To be maybe a bit clearer, ``` `sameValue` treats "NaN"^^xsd:double as the "same value" as _xyz_.```

spec/index.html
afs: It is now a note.

Tpt: thanks!

spec/index.html
afs: Incorporated the first part of @rubensworks suggestion : "is defined to return"

The latter part revised to 

```
                However, `sameTerm("NaN"^^xsd:double, "NaN"^^xsd:double)` is true.
                The function `sameValue` defines `sameValue("NaN"^^xsd:double, "NaN"^^xsd:double)`
                to be true because the arguments are the same element of the value space.
```

afs:
@kasei , @hartig 

Is the current state of this PR satisfactory for merge? If so, could you approve it?

hartig:
> Is the current state of this PR satisfactory for merge? If so, could you approve it?

Andy, sorry for having been quiet in this PR recently. I have an important deadline tomorrow and will be able to come back to this PR only later this week.

kasei:
> @kasei , @hartig 
> 
> Is the current state of this PR satisfactory for merge? If so, could you approve it?

I'll take a look today.

kasei:
This looks good to me with one exception: the special handling for "NaN" in §17.4.2.2 sameValue. The content in §17.4.2.2 seems reasonable, and the rationale provided in the following NOTE section is good. However, I'm not sure how you'd ever get into `sameValue` when trying to evaluate something like `"NaN"^^xsd:double = "NaN"^^xsd:float` because §17.3 Operator Mapping seems clear that the rule for numeric types (using `op:numeric-equal`) would take precedence over the fallback rule for RDF terms (using `sameValue`).

Since this is value-specific, we can't simply add a new row to the operator mapping table, but I think something needs to be done to ensure handling of NaN values ends up using `sameValue` semantics rather than `numeric-equal` semantics.

kasei:
Sorry. I hadn't been able to follow all the discussion in the thread. Just to be clear - since `sameValue` isn't exposed as a callable function, is there any practical use to this definition (w.r.t. "NaN" values)? Or is it just to align with the conceptual need for "sameTerm implies sameValue"? I *think* it's the latter based on the comments above from @Tpt (?). If that's the case, I'm OK with this being merged, though I think it might be a point of confusion to readers of the spec in the future.

kasei:
> Anything to do with NaNs is confusing in the details. Intuitions mislead.

Agreed. My concern is more around the fact that the document is defining behavior that can't ever be used, though (so not really about NaNs… could be about anything). I'm happy to let this get merged, but it strikes me as really strange to be adding definitions that can't possibly be used/observed in a SPARQL system. Obviously it would be different if `sameValue` were callable.

afs:
Issue #201 for whether `sameValue` should be callable.

hartig (approved):

afs:
Squashed authoring, editing and review commits.

spec/index.html
```diff
@@ -5669,248 +5731,222 @@ <h5>NOT IN</h5>
...
-                and both literals have the same <a data-cite="RDF12-CONCEPTS#dfn-literal-value">literal value</a>.
+                `"NaN"^^xsd:double` and `"NaN"^^xsd:float` are considered to 
+                represent the same value.
+                If <code>term1</code> and <code>term2</code> are
+                both "NaN" for either xsd:double or xsd:float, then 
+                return TRUE.
+              </li>
+              <li>If <code>term1</code> and <code>term2</code> are both 
+                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
+                and the SPARQL processor can determine that their the values are equal,
```
TallTed, line 5857:
```suggestion
                and the SPARQL processor can determine that their values are equal,
```

spec/index.html
```diff
@@ -5669,248 +5731,222 @@ <h5>NOT IN</h5>
...
+                then return TRUE.
+              </li>
+              <li>If <code>term1</code> and <code>term2</code> are both 
+                <a data-cite="RDF12-CONCEPTS#dfn-literal">literals</a>
+                and the SPARQL processor can determine the values can not be equal,
```
TallTed, line 5862:
```suggestion
                and the SPARQL processor can determine that their values cannot be equal,
```

spec/index.html
```diff
@@ -5669,248 +5731,222 @@ <h5>NOT IN</h5>
...
+                if its datatype is handled by the SPARQL processor and
+                its lexical form is not in the
+                <a data-cite="RDF12-CONCEPTS#dfn-lexical-space">lexical space</a>
+                of the datatype.
+              </p>
+              <p>
+                If the two arguments are literals, the function `sameValue` 
+                returns `true` or `false` in cases where the SPARQL processor
+                can determine that the values of these literals are equal or are not equal. 
+                If the SPARQL processor can not be sure, it returns `error`.
```
TallTed, line 5887:
```suggestion
                can determine that the values of these literals are equal or are not equal, respectively.
                If the SPARQL processor cannot be sure, it returns `error`.
```

Issue #182, opened 2025-01-22
Cover triple terms in definition of  RDFterm-equal


hartig:
By the current definition of [RDFterm-equal](https://w3c.github.io/sparql-query/spec/#func-RDFterm-equal), comparing two triple terms based on this function (i.e., using the `=` operator) always returns `false`, even if the two triple terms are the same.

The definition needs to be extended to consider triple terms in the second list of bullet points (the one that specifies what it means for two RDF terms to be _equal_).

hartig:
Related to this issue, see [Section 4.4.9](https://www.w3.org/2021/12/rdf-star.html#rdf-star-termEqual) of the CG report.

PR #197, opened 2025-02-18
Renames 'RDFterm-equal' to 'sameValue'

### LINKED ISSUE #25: Rename 'RDFterm-equal'

hartig:
As discussed in #25, renames the `RDFterm-equal` function to `sameValue`. Also, [as suggested](https://github.com/w3c/sparql-query/issues/25#issuecomment-2610386674) by @afs, this PR adds a legacy anchor (with an HTML comment that this is anchor is obsolete now, as done for similar cases in RDF-Concepts)

For the sake of keeping this PR small and focused, it does *not* include any changes to the definition of the renamed function (as are currently discussed in #187, and also related to #182).

Closes #25

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/197.html" title="Last updated on Feb 18, 2025, 2:08 PM UTC (73f6d08)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/197/1dabcb9...73f6d08.html" title="Last updated on Feb 18, 2025, 2:08 PM UTC (73f6d08)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -5280,7 +5280,7 @@ WHERE { ?annot  a:annotates  &lt;http://www.w3.org/TR/rdf-sparql-query/&gt; .
               <td><span class="type RDFterm">RDF term</span></td>
               <td><span class="type RDFterm">RDF term</span></td>
               <td class="xpathOp">
-                <a href="#func-RDFterm-equal" class="SPARQLoperator">RDFterm-equal</a>(A, B)
+                <a href="#func-sameValue" class="SPARQLoperator">sameValue</a>(A, B)
               </td>
               <td>xsd:boolean</td>
             </tr>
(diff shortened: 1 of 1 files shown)


afs:
This will conflict with #194 which reorders the sections.

hartig:
> This will conflict with #194 which reorders the sections.

Ah! I didn't pay attention to #194, probably because it is marked as draft. Sorry!!

In this case, I will keep this PR open for the moment but are ready to drop it if we decide to go forward with #194 (for which I would have preferred it didn't contain both the change of the order of the sections and the change of the definition of `RDFterm-equal`).

Tpt (approved):

afs:
I don't think we have decided whether to have `sameValue` in the grammar or not.

afs:
> (for which I would have preferred it didn't contain both the change of the order of the sections and the change of the definition of `RDFterm-equal`).

Sorry, but there are overheads when there are chains PRs.

hartig:
> I don't think we have decided whether to have `sameValue` in the grammar or not.

The change in this PR is not about adding anything to the grammar but only about changing the name of the function that is used to define `=` in the last two cases in the operator mapping table. I know that [you also proposed](https://github.com/w3c/sparql-query/issues/25#issuecomment-1470284336) to consider supporting this function directly as a new keyword, but as I saw this as an orthogonal question, I didn't include it in this PR.

hartig:
> Sorry, but there are overheads when there are chains PRs.

Understood.

rubensworks (approved):

afs:
After trying to apply this to #194 and finding the patch does not apply - and there new uses of RDFterm-equal in #194 - 
I've copied changes into #194.

There are now two uses of the string `RDFterm-equal` - the obsolate anchor and the "Changes" entry.

afs:
This is carried out in #194 which is in the same area.

PR #216, opened 2025-05-09
Linked function names of algebra operators, and markup of variables in Section 18.5, 18.5.1, and 18.5.2

hartig:
One thing that has always bugged me is that the spec represents the operators of the abstract/algebraic syntax of SPARQL in exactly the same way as it represents the algebra functions used to define the evaluation semantics. As an example, consider the following formula from Section [18.5.2 Evaluation Semantics](https://www.w3.org/TR/sparql12-query/#sparqlAlgebraEval).

> eval(D(G), Filter(F, P)) = Filter(F, eval(D(G),P), D(G))

The 'Filter' on the left-hand side of this formula represents an operator of the algebraic syntax (as introduced in the second table of Section [18.2 Translation to the SPARQL Algebra](https://www.w3.org/TR/sparql12-query/#sparqlQuery)), whereas the 'Filter' on the right-hand side refers to the algebra operator defined in Section [18.5 SPARQL Algebra](https://www.w3.org/TR/sparql12-query/#sparqlAlgebra). As you can see, they look identical.

This has often been a source of confusion for me. While the formulas of the form as above become understandable once one has figured it out (which is not trivial for someone new coming to the spec), it is worse for cases in which something like 'Filter( ... )' is mentioned elsewhere, in particular just within text.

This PR is a first step towards fixing this issue. All mentions of the algebra operators are wrapped as links to their corresponding definition in Section [18.5 SPARQL Algebra](https://www.w3.org/TR/sparql12-query/#sparqlAlgebra), and these links have a dedicated CSS `class` attribute that allows us to make them visually different in the future. Additionally, the PR adds another such `class` attribute for the (earlier-introduced) links to the set functions, and it adds `<var>..</var>` markup for all formulas in Sections [18.5 SPARQL Algebra](https://www.w3.org/TR/sparql12-query/#sparqlAlgebra), [18.5.1 Aggregate Algebra](https://www.w3.org/TR/sparql12-query/#aggregateAlgebra), and [18.5.2 Evaluation Semantics](https://www.w3.org/TR/sparql12-query/#sparqlAlgebraEval). The example formula above looks as follows now.

> [eval](https://www.w3.org/TR/sparql12-query/#sparqlAlgebraEval)( _D_(_G_), Filter(_F_, _P_) ) = [Filter](https://www.w3.org/TR/sparql12-query/#defn_algFilter)( _F_, [eval](https://www.w3.org/TR/sparql12-query/#sparqlAlgebraEval)(_D_(_G_), _P_), _D_(_G_) )

I am leaving it to a future PR to do something similar for the operators of the algebraic syntax (e.g., the 'Filter' on the left-hand side of this formula) in order to avoid making this PR too complex. For the same reason, this PR does not change any of the actual content.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -8374,10 +8374,10 @@ WHERE {
             <p>A <span class="definedTerm">solution sequence modifier</span> is one of:</p>
             <ul>
               <li>
-                <a href="#defn_algOrdered">Order By</a> modifier: put the solutions in order
+                <a href="#defn_algOrderBy">Order By</a> modifier: put the solutions in order
               </li>
               <li>
-                <a href="#defn_algProjection">Projection</a> modifier: choose certain variables
+                <a href="#defn_algProject">Projection</a> modifier: choose certain variables
               </li>
               <li>
                 <a href="#defn_algDistinct">Distinct</a> modifier: ensure solutions in the sequence
@@ -9759,11 +9759,11 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
           described above.</p>
         <div class="defn">
           <p><b>Definition: <span id="defn_algFilter">Filter</span></b></p>
-          <p>Let Ω be a multiset of solution mappings and expr be an expression. We define:</p>
-          <p>Filter(expr, Ω) = { μ | μ in Ω and expr(μ) is an expression that has an
+          <p>Let <var>Ω</var> be a multiset of solution mappings and <var>expr</var> be an expression. We define:</p>
+          <p><a href="#defn_algFilter" class="algFct">Filter</a>(<var>expr</var>, <var>Ω</var>) = { <var>μ</var> | <var>μ</var> in <var>Ω</var> and <var>expr</var>(<var>μ</var>) is an expression that has an
             effective boolean value of true }</p>
-          <p><a href="#defn_Multiplicity">multiplicity</a>( μ | Filter(expr, Ω) )
-            = <a href="#defn_Multiplicity">multiplicity</a>( μ | Ω )</p>
+          <p><a href="#defn_Multiplicity">multiplicity</a>( <var>μ</var> | <a href="#defn_algFilter" class="algFct">Filter</a>(<var>expr</var>, <var>Ω</var>) )
+            = <a href="#defn_Multiplicity">multiplicity</a>( <var>μ</var> | <var>Ω</var> )</p>
           <blockquote>
             Note that evaluating an <code>exists(pattern)</code> expression uses the dataset and
             active graph, D(G). See the <a href="#defn_evalFilter">evaluation of filter</a>.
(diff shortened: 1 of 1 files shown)


domel (approved):

TallTed:
I will resist the temptation to read the actual content and revise for comprehension. :-)

I look forward to the near-future PR which does "something similar for the operators of the algebraic syntax", which actual content I will also ignore.

Then I will make my own review of the actual content of Sections [18.5 SPARQL Algebra](https://www.w3.org/TR/sparql12-query/#sparqlAlgebra), [18.5.1 Aggregate Algebra](https://www.w3.org/TR/sparql12-query/#aggregateAlgebra), and [18.5.2 Evaluation Semantics](https://www.w3.org/TR/sparql12-query/#sparqlAlgebraEval).

hartig:
@TallTed I think there is plenty to revise. I have a bunch of ideas for improving the formal parts, but my intention is to keep this separate from pure markup improvements to keep the PRs focused.

Tpt (approved):
Thank you so much for this work.

afs (approved):
Looks great and a good improvement.

I do find the linking "eval" as less useful because it occurs everywhere, several times, to the point of being a distracting.

However, it is the same HTML in the document. Possibly we could style it differently (e.g. no underline) at some point.

hartig:
@afs I admit, I find the underlining of the now-linked (as per this PR) function names a bit ugly and distracting as well. The explicit CSS classes introduced by this PR do indeed allow us to style these particular links differently, removing the underlining.

Having said that, linking all the occurrences of the eval function to the start of the recursive definition of this function is useful, I think; in particular, because there is another function called "eval" in the spec, namely the one for Property Path patterns in Section [18.4 Property Path Patterns](https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/216.html#PropertyPathPatterns).

rubensworks (approved):

spec/index.html
```diff
@@ -9759,153 +9759,149 @@ <h3>SPARQL Algebra</h3>
...
-            <a href="#defn_Multiplicity">multiplicity</a>( μ | Ω )</p>
+          <p>Let <var>Ω</var> be a multiset of solution mappings. We define:</p>
+          <p><a href="#defn_algToList" class="algFct">ToList</a>(<var>Ω</var>) = a sequence of mappings <var>μ</var> in <var>Ω</var> in any order, with
+            <a href="#defn_Multiplicity">multiplicity</a>( <var>μ</var> | <var>Ω</var> ) occurrences of
+            <var>μ</var></p>
+          <p><a href="#defn_Multiplicity">multiplicity</a>( <var>μ</var> | <a href="#defn_algToList" class="algFct">ToList</a>(<var>Ω</var>) ) =
+            <a href="#defn_Multiplicity">multiplicity</a>( <var>μ</var> | <var>Ω</var> )</p>
         </div>
         <div class="defn">
-          <p><b>Definition: <span id="defn_algOrdered">OrderBy</span></b></p>
```
kasei, line 9857: We seem to be losing existing anchors in this change. Could we preserve `id` values in adjacent markup (the enclosing `<p>` or `<b>` perhaps) to allow any existing deep links into the 1.1 spec to still work with the 1.2 spec?

afs: Elsewhere, and in other docs, there are empty span's e.g.`<span id=><!-- comment --></span>` or `<span id=/>`.

TallTed:
> there is another function called "eval" in the spec

This makes me think that one of these functions should be renamed, and if any other names are used for multiple functions, that all but one of each such group should be renamed.

afs:
It would be possible to call the other one "evalPath".

hartig:
> > > @hartig: there is another function called "eval" in the spec
> >
> > @TallTed: This makes me think that one of these functions should be renamed [...]
>
> @afs It would be possible to call the other one "evalPath".

I have created a separate issue for this, #218

spec/index.html
hartig: > We seem to be losing existing anchors in this change. [...]

I checked that these anchors are not used anywhere in the spec, but you are right, someone else may have used them externally. I have restored them now (commit https://github.com/w3c/sparql-query/pull/216/commits/38052ef109ffd3c9e7ec7d9fe32bb9f4da67e90a) using empty `<span>` elements as we have already done elsewhere for other obsolete anchors.

hartig: @kasei with the anchors restored now, do you approve this PR?

kasei (approved):

Issue #217, opened 2025-05-10
Evaluation of GRAPH pattern ignores blank nodes as graph names


hartig:
The [definition of how to evaluate GRAPH patterns](https://www.w3.org/TR/sparql12-query/#defn_evalGraph) uses the following pseudo code for the case of GRAPH patterns with a variable rather than an IRI.
```
eval(D(G), Graph(var,P)) =
    Let R be the empty multiset
    foreach IRI i in D
        R := Union(R, Join( eval(D(D[i]), P) , Ω(?var->i) ) )
    the result is R
```
The `foreach` loop in the third line considers only IRIs as graph names in the queried dataset. Yet, given that we now have adopted the RDF Concepts notion of RDF dataset also for SPARQL, graph names may also be blank nodes. The current definition GRAPH pattern evaluation ignores this option.

We can easily fix this issue by extending the third line of the pseudo code above such that it considers blank nodes as well, but this requires also extending the definition of the notation `D[i]` as used in the fourth line. Additionally, it would be useful to make more explicit that the third line actually refers to the graph names in `D`.

Implementing this fix has to wait until after PR #216 has been merged.

PR #221, opened 2025-05-17
Consider bnode graph names in evaluation of Graph

hartig:
This PR addresses #217 by extending the [definition of how to evaluate GRAPH patterns](https://www.w3.org/TR/sparql12-query/#defn_evalGraph) such that it now also considers blank nodes as possible graph names in an RDF dataset.

Additionally, the PR:
* changes the layout style of the changed definition such that it is consistent with the style of the other definitions in this part of the spec
* removes the introduction of the notation _Ω_(_x_) from the beginning of Section [18.3 Basic Graph Patterns](https://www.w3.org/TR/sparql12-query/#BasicGraphPattern)
  * (this notation was used only in the definition that this PR changes and it was also not really clear; I have replaced it by something better directly in the changed definition now)
* removes the notation _D_[_i_] from the beginning of Section [18.5.2 Evaluation Semantics](https://www.w3.org/TR/sparql12-query/#sparqlAlgebraEval)
  * (this notation was also used only in the definition that this PR changes)
* removes the following paragraph which was placed directly below the changed definition but, in my opinion, doesn't really add anything (in fact, it is not even relevant because there can be no solution mapping that is in the result of more than one of the joins that feed into the union in the given algorithm)---but if you think it is useful to keep that paragraph, let me know and I will restore it!
> The evaluation of graph uses the [Union](https://www.w3.org/TR/sparql12-query/#defn_algUnion) operator. The multiplicity of a solution mapping is the sum of the multiplicities of that solution mapping in each [Join](https://www.w3.org/TR/sparql12-query/#defn_algJoin) operation.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/221.html" title="Last updated on Jun 5, 2025, 8:13 AM UTC (7646624)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/221/5615dfa...7646624.html" title="Last updated on Jun 5, 2025, 8:13 AM UTC (7646624)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -10373,7 +10373,7 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
                 where <var>Ω</var> is the multiset of solution mappings produced by the following algorithm:
                 <pre class="code nohighlight" style="font-family: sans-serif;">
 <var>Ω</var> := the empty multiset
-foreach <a data-cite="RDF12-CONCEPTS#dfn-graph-name">graph name</a> <var>gn</var> in <var>D</var>
+foreach <a data-cite="RDF12-CONCEPTS#dfn-graph-name">graph name</a> <var>gn</var> in <var>D</var> (recall that a graph name may be an IRI or a blank node)
     <var>G'</var> := the RDF graph of the named graph with name <var>gn</var> in <var>D</var>
     <var>Ω'</var> := <a href="#defn_eval" class="evalFct">eval</a>( <var>D</var>(<var>G'</var>), <var>P</var> )
     <var>Ω</var> := <a href="#defn_algUnion" class="algFct">Union</a>( <var>Ω</var>, <a href="#defn_algJoin" class="algFct">Join</a>(<var>Ω'</var>, <var>μ</var>) ), where <var>μ</var> = {<var>x</var> → <var>gn</var>}
--- a/spec/index.html
+++ b/spec/index.html
@@ -107,6 +107,7 @@ pre.prototype	{ background-color:#f7f8ff;
 div.defn p	{ margin-top: 1ex ; margin-bottom: 1.5ex ;}
 div.defn ul	{ margin-top: 1ex ; margin-bottom: 1.5ex ; }
 span.definedTerm	{font-weight: bold;}
+div.indentedFormula { margin-left: 5ex ; margin-top: 2mm ; margin-bottom: 2mm ; }
 
 div.grammarExtract
                 { border: thin solid #888888;
@@ -9448,8 +9449,6 @@ The set PV is used later for projection.
         <p>Write <var>Ω<sub>0</sub></var> for the multiset consisting of exactly the empty mapping
           <var>μ<sub>0</sub></var>, with multiplicity 1. This is the join identity.</p>
         <p>Write <var>μ</var>(<var>x</var>) for the solution mapping variable <var>x</var> to RDF term <var>t</var> : { (<var>x</var>, <var>t</var>) }.</p>
-        <p>Write <var>Ω</var>(<var>x</var>) for the multiset consisting of exactly <var>μ</var>(<var>?x</var>-&gt;<var>t</var>), that is, <code>{ { (x, t) } }</code>
-          with multiplicity 1.</p>
         <div class="defn">
           <p><b>Definition: <span id="defn_algCompatibleMapping">Compatible Mappings</span></b></p>
           <p>Two solution mappings μ<sub>1</sub> and μ<sub>2</sub> are compatible if, for every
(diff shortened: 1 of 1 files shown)


kasei (approved):

Tpt (approved):

spec/index.html
```diff
@@ -10344,25 +10342,46 @@ <h3>Evaluation Semantics</h3>
...
-            </pre>
-            <pre class="code nohighlight">
-<a href="#defn_eval" class="evalFct">eval</a>( <var>D</var>(<var>G</var>), Graph(<var>var</var>,<var>P</var>) ) =
-    Let <var>R</var> be the empty multiset
-    foreach IRI <var>i</var> in <var>D</var>
-        <var>R</var> := <a href="#defn_algUnion" class="algFct">Union</a>(<var>R</var>, <a href="#defn_algJoin" class="algFct">Join</a>( <a href="#defn_eval" class="evalFct">eval</a>(<var>D</var>(<var>D</var>[<var>i</var>]), <var>P</var>) , Ω(<var>var</var>-&gt;<var>i</var>) ) )
-    the result is <var>R</var>
-            </pre>
+            <p>For every |x| that is
+              an <a data-cite="RDF12-CONCEPTS#dfn-IRI">IRI</a> or
```
afs, line 10346:
```suggestion
              an <a data-cite="RDF12-CONCEPTS#dfn-IRI">IRI</a>,
              a <a data-cite="RDF12-CONCEPTS#dfn-blank-node">blank node</a> or
```

Should the definition of the evaluation of "graph" mention blank nodes?

It isn't necessary because they come in via variables but:

1. It more clearly indicates that literal and triple terms can not be used
2. The intuitive of variable substitution applies

hartig: @afs your edit suggestion is incorrect. The _x_ in Graph(_x_, _P_) may indeed be only an IRI or a variable; not a blank node. While the algebraic syntax is not explicitly defined in the spec, it is clear from the grammar of the user-facing syntax that there can be no blank node---see the [`GraphGraphPattern`production](https://www.w3.org/TR/sparql12-query/#rGraphGraphPattern):
```
[62] GraphGraphPattern ::= 'GRAPH' VarOrIri GroupGraphPattern
```
> Should the definition of the evaluation of "graph" mention blank nodes?
>
> It isn't necessary because they come in via variables but:
>
> 1. It more clearly indicates that literal and triple terms can not be used

Cannot be used for what exactly?

If you mean that they cannot be used as the _x_ in the new version of this definition, I think stating that _x_ may be an IRI or a variable (as done in this PR) is sufficient.

If you mean that they cannot be used as bindings for the variable _x_ (in case _x_ is a variable), I don't think that saying something like this belongs in this definition; maybe better suited for the informal text in Section [13.3 Querying the Dataset](https://www.w3.org/TR/sparql12-query/#queryDataset). Also, it is not entirely correct to say that literals and term maps cannot be assigned to variable _x_ in incoming solution mappings; such solution mappings would just not survive the Join within the foreach loop. Perhaps one thing to make this more obvious would be to extend the foreach-line of the algorithm from (current form)

foreach [graph name](https://www.w3.org/TR/rdf12-concepts/#dfn-graph-name) _gn_ in _D_

to

foreach [graph name](https://www.w3.org/TR/rdf12-concepts/#dfn-graph-name) _gn_ in _D_ (recall that a graph name may be an IRI or a blank node)

Would that address your concern?

> 2. The intuition of variable substitution applies

I am not sure what you mean by this comment.

afs: > >1. It more clearly indicates that literal and triple terms can not be used

> Cannot be used for what exactly?

As graph names.

> > 2. The intuition of variable substitution applies

> I am not sure what you mean by this comment.

In TriG, `GRAPH _:label { ... }` is legal so one can think of the query as a trig pattern, like BGPs are turtle patterns.

The variable may, visually, be defined in the query close-by.

(Yes, this actually happens when the "graph" evaluation is joined with the rest of the group graph pattern. Optimizers are prone to use known graph name though.)

hartig: @afs 
>>> 1. It more clearly indicates that literal and triple terms can not be used
>>
>>    Cannot be used for what exactly?
>
> As graph names.
>
> It does not cause any harm to be a little bit more general here.
> The 1.1 text says "foreach IRI i in D"; the new text puts in a link so the cue isn't there.
> 
> "foreach [graph name](https://www.w3.org/TR/rdf12-concepts/#dfn-graph-name) gn in D"

Now I understand your concern! I had a proposal related to that already in my previous comment: We may extend this line of the algorithm as follows.

"foreach [graph name](https://www.w3.org/TR/rdf12-concepts/#dfn-graph-name) _gn_ in _D_ (recall that a graph name may be an IRI or a blank node)"

I think that should address your concern. Do you agree?

>>> 2. The intuition of variable substitution applies
>>
>> I am not sure what you mean by this comment.
>
> In TriG, `GRAPH _:label { ... }` is legal so one can think of the query as a trig pattern, like BGPs are turtle patterns.

Right.

Do you think my proposal for the previous point addresses this one as well? Otherwise, what else would you propose to do here?

rubensworks (approved):

kasei (approved):

spec/index.html
TallTed:
```suggestion
              an <a data-cite="RDF12-CONCEPTS#dfn-IRI">IRI</a>, or
```

spec/index.html
```diff
@@ -10373,7 +10373,7 @@ <h3>Evaluation Semantics</h3>
                 where <var>Ω</var> is the multiset of solution mappings produced by the following algorithm:
                 <pre class="code nohighlight" style="font-family: sans-serif;">
 <var>Ω</var> := the empty multiset
-foreach <a data-cite="RDF12-CONCEPTS#dfn-graph-name">graph name</a> <var>gn</var> in <var>D</var>
+foreach <a data-cite="RDF12-CONCEPTS#dfn-graph-name">graph name</a> <var>gn</var> in <var>D</var> (recall that a graph name may be an IRI or a blank node)
```
TallTed, line 10376: `foreach` is found in various programming languages. `for each` is the English phrase.
```suggestion
for each <a data-cite="RDF12-CONCEPTS#dfn-graph-name">graph name</a> <var>gn</var> in <var>D</var> (recall that a graph name may be an IRI or a blank node)
```

spec/index.html
afs: Just noting that `GRAPH _:blank` (if it were legal) is always empty. Blank nodes treatment as non-distinguished variables happens in BGPs.

afs (approved):

Issue #249, opened 2025-07-12
bug in translation when gathering FILTER expresssions


pfps:
```18.3.2.2 Collect FILTER Elements

FILTER expressions apply to the whole group graph pattern in which they appear. The algebra operators to perform filtering are added to the group after translation of each group element. We collect the filters together here and remove them from group, then [apply them to the whole translated group graph pattern](https://www.w3.org/TR/sparql12-query/#sparqlAddFilters).

In this step, we also translate graph patterns within FILTER expressions [EXISTS](https://www.w3.org/TR/sparql12-query/#func-filter-exists) and [NOT EXISTS](https://www.w3.org/TR/sparql12-query/#func-filter-not-exists).

Let FS := empty set
For each form FILTER(expr) in the group graph pattern
    In expr, replace NOT EXISTS{P} with fn:not(exists(translate(P))) 
    In expr, replace EXISTS{P} with exists(translate(P))
    FS := FS ∪ {expr}
    End

The set of filter expressions FS is [used later](https://www.w3.org/TR/sparql12-query/#sparqlAddFilters).
```

There is no definition of that in means here.    

For example, why is the filter in the query below

```
SELECT ?x WHERE {
  ?x :a :b .
 MINUS { ?x :c :d . FILTER (?x = :w) }
}
```
not in the main group graph pattern of the SELECT?

afs:
https://www.w3.org/TR/sparql12-query/#convertGraphPattern

The conversion to the algebra applies to Group Graph Patterns.

It is applied in nested fashion to the RHS of `MINUS` at

```
    If E is of the form MINUS{P}
        G := Minus(G, Translate(P))
        End
```

pfps:
See #250

It was better to roll the processing of FILTER into the processing of groupgraphpatternsub.

PR #250, opened 2025-07-14
better handling of FILTER in translation to algebra

pfps:
The current processing of FILTER expressions depends on an ill-defined "in" predicate and an unclear notion of what descendants of a groupgraphpattern are processed in its function.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -8924,28 +8924,6 @@ WHERE {
             <p>Expand abbreviations for IRIs and triple patterns given in 
               <a href="#sparqlSyntax">section 4</a>.</p>
           </section>
-          <section id="sparqlCollectFilters">
-            <h5>Collect <code>FILTER</code> Elements</h5>
-            <p><code>FILTER</code> expressions apply to the whole group graph pattern in which they
-              appear. The algebra operators to perform filtering are added to the group after
-              translation of each group element. We collect the filters together here and remove them
-              from group, then <a href="#sparqlAddFilters">apply them to the whole translated group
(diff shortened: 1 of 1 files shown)


hartig:
I very much agree with the observation in #249 that the part about collecting the FILTER elements can be improved. Especially, the scope of this part, in combination with its integration into the overall translation algorithm should be made more explicit. In this sense, I generally like the direction of this PR.

However, simply moving the FILTER collection into the 'Translate Graph Patterns' section as done by this PR changes the outcome of the translation algorithm in terms of the basic graph patterns (BGPs) that are sub-expressions of the resulting algebraic expression. That is, originally (i.e., without this PR), FILTER elements are collected **and removed** before subsequent triple patterns are combined into BGPs (including also triple patterns created when rewriting property path patterns). Given the change in this PR, triple patterns are combined into BGPs already before removing the collected FILTER elements.

As a concrete example, consider the following group graph pattern.
```sparql
{
  ?x :p ?y
  FILTER ( isIRI(?y) )
  ?y :q ?z
}
```

The original translation algorithm converts this into a Filter over a single BGP that consists of both triple patterns. In contrast, the translation algorithm in this PR converts this into a Filter over a Join of two BGPs, where each of these BGPs contains one of the two triple patterns, respectively.

Having said that, perhaps this change is not such a big deal because it doesn't have an impact on the query result (a join of two BGPs is semantically equivalent to a single BGP that is the union of the other two BGPs).

(Something else: this PR conflicts with PR #245, in the sense of one causing a merge conflict with the other.)

Issue #254, opened 2025-08-13
Expressions and their evaluation are ill-defined


hartig:
While working on a PR for #212, I notice that the notion of an [expression](https://www.w3.org/TR/sparql12-query/#expressions) and the evaluation of such expressions is not really defined in the spec.

### What is an expression?
Section [17 Expressions and Testing Values](https://www.w3.org/TR/sparql12-query/#expressions) mentions "the expression" in the second sentence without saying what kind of expression this refers to.

A few paragraphs later, we have the following:
> SPARQL expressions are constructed according to the grammar and provide access to functions [...]

Assuming that "SPARQL expression" and "expression" are the same concept, my takeaway from this sentence is that a (SPARQL) expression is a string admitted by some particular production of the SPARQL grammar. Okay, that's fine. Or, is it actually meant to be an AST obtained by parsing a such string based on the grammar??

In any case, it would be useful to make this an explicit definition, and also to say explicitly which production of the SPARQL grammar is the one that is meant here (is it [`Expression`](https://www.w3.org/TR/sparql12-query/#rExpression) as used in, e.g., [`Bind`](https://www.w3.org/TR/sparql12-query/#rBind) or is it [`Constraint`](https://www.w3.org/TR/sparql12-query/#rConstraint) as used in [`Filter`](https://www.w3.org/TR/sparql12-query/#rFilter)?)

### How is the evaluation of an expression defined?
Now, this one is a bit more tricky. There is Section [17.2 Filter Evaluation](https://www.w3.org/TR/sparql12-query/#evaluation), which has informal text that covers various cases but there is nothing that explicitly says something like: "Given an expression _expr_, the result of evaluating _expr_ is ..." In other words, we are missing something akin to the [eval](https://www.w3.org/TR/sparql12-query/#defn_eval) function but for [expressions](https://www.w3.org/TR/sparql12-query/#expressions).

Another aspect of the evaluation of expressions that is not mentioned at all in Section [17.2 Filter Evaluation](https://www.w3.org/TR/sparql12-query/#evaluation) is that an expression is not evaluated per se, but with respect to a solution mapping (and, in fact, also with respect to a dataset with an active graph, if we want to be able to properly define the semantics of expressions with EXISTS). The only part of Section [17 Expressions and Testing Values](https://www.w3.org/TR/sparql12-query/#expressions) related to the fact that solution mappings are considered for the evaluation of expressions is the second sentence, which says that
[...]

Issue #212, opened 2025-05-03
Inconsistency in the definition versus the use of the 'Filter' algebra operator


hartig:
The definition of [Evaluation of Filter](https://w3c.github.io/sparql-query/spec/#defn_evalFilter) in [Sec.18.5.2 (Evaluation Semantics)](https://w3c.github.io/sparql-query/spec/#sparqlAlgebraEval) currently says:
```
eval(D(G), Filter(F, P)) = Filter(F, eval(D(G),P), D(G))
```
Notice that the function on the right-hand side of this equation is invoked with _**three**_ arguments:
i) `F` (the filter expression),
ii) `eval(D(G),P)` (the multiset of solution mappings that results from evaluating the child pattern `P`), and
iii) `D(G)` (the dataset and active graph).

Yet, when looking at [the definition of this function](https://w3c.github.io/sparql-query/spec/#defn_algFilter),
```
Let Ω be a multiset of solution mappings and expr be an expression. We define:
Filter(expr, Ω) = ...
```
it turns out that the function is defined for only **_two_** arguments:
i) a filter expression and
ii) a multiset of solution mappings.

This inconsistency needs to be fixed.

By the way, it was not like that in SPARQL 1.0. [Back then, the Evaluation of Filter](https://www.w3.org/TR/sparql10-query/#defn_evalFilter) used the [Filter function](https://www.w3.org/TR/sparql10-query/#defn_algFilter) with its two arguments.

I assume that the reason for passing the active graph as a third argument is related to the idea that the [evaluation](https://w3c.github.io/sparql-query/spec/#defn_evalExists) of [Exists](https://w3c.github.io/sparql-query/spec/#defn_exists) requires access to the active graph.

Fixing this issue properly is not entirely trivial I am afraid. All related definitions need to be extended such that `D(G)` can be passed as an additional argument, including the definition of how expressions are evaluated.

PR #255, opened 2025-08-13
Fixes inconsistency in the definition versus the use of the 'Filter' algebra operator

hartig:
The main change in this PR, which addresses #212, is to extend the signature of the [Filter](https://www.w3.org/TR/sparql12-query/#defn_algFilter) algebra operator from [Filter](https://www.w3.org/TR/sparql12-query/#defn_algFilter)(_expr_, _Ω_) to [Filter](https://www.w3.org/TR/sparql12-query/#defn_algFilter)(_expr_, _Ω_, _D_, _G_), where _D_ is a dataset and _G_ is the active graph.

Additionally, the PR adds notes about issue #254.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -9904,14 +9904,20 @@ Let x and y be variables or RDF terms, and S a set of IRIs:
           described above.</p>
         <div class="defn">
           <p><b>Definition: <span id="defn_algFilter">Filter</span></b></p>
-          <p>Let <var>Ω</var> be a multiset of solution mappings and <var>expr</var> be an expression. We define:</p>
-          <p><a href="#defn_algFilter" class="algFct">Filter</a>(<var>expr</var>, <var>Ω</var>) = { <var>μ</var> | <var>μ</var> in <var>Ω</var> and <var>expr</var>(<var>μ</var>) is an expression that has an
+          <p>Let <var>Ω</var> be a multiset of solution mappings,
+            <var>expr</var> be an <a href="#expressions">expression</a>,
+            |D| be a <a href="#sparqlDataset">dataset</a>,
+            and |G| be the <a href="#defn_ActiveGraph">active graph</a>.
+            We define:</p>
+          <p><a href="#defn_algFilter" class="algFct">Filter</a>(<var>expr</var>, <var>Ω</var>, |D|, |G|) = { <var>μ</var> in <var>Ω</var>  | <var>expr</var>(<var>μ</var>) is an expression that has an
             effective boolean value of true }</p>
-          <p><a href="#defn_Multiplicity">multiplicity</a>( <var>μ</var> | <a href="#defn_algFilter" class="algFct">Filter</a>(<var>expr</var>, <var>Ω</var>) )
+<div class="issue" data-number="254">
+  It is not clear what <var>expr</var>(<var>μ</var>) is, and it is not apparent in the formula that the expression |expr| is meant to be evaluated not only with respect to <var>μ</var> but also with respect to |D| with active graph |G|.</div>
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -9904,14 +9904,20 @@ <h3>SPARQL Algebra</h3>
...
+            |D| be a <a href="#sparqlDataset">dataset</a>,
+            and |G| be the <a href="#defn_ActiveGraph">active graph</a>.
+            We define:</p>
+          <p><a href="#defn_algFilter" class="algFct">Filter</a>(<var>expr</var>, <var>Ω</var>, |D|, |G|) = { <var>μ</var> in <var>Ω</var>  | <var>expr</var>(<var>μ</var>) is an expression that has an
             effective boolean value of true }</p>
-          <p><a href="#defn_Multiplicity">multiplicity</a>( <var>μ</var> | <a href="#defn_algFilter" class="algFct">Filter</a>(<var>expr</var>, <var>Ω</var>) )
+<div class="issue" data-number="254">
+  It is not clear what <var>expr</var>(<var>μ</var>) is, and it is not apparent in the formula that the expression |expr| is meant to be evaluated not only with respect to <var>μ</var> but also with respect to |D| with active graph |G|.</div>
+          <p><a href="#defn_Multiplicity">multiplicity</a>( <var>μ</var> | <a href="#defn_algFilter" class="algFct">Filter</a>(<var>expr</var>, <var>Ω</var>, |D|, |G|) )
             = <a href="#defn_Multiplicity">multiplicity</a>( <var>μ</var> | <var>Ω</var> )</p>
```
TallTed, line 9917: It appears likely that these `<p>` should instead be a list (I think unordered, but I could be wrong) beneath the `We define:` which I suggest should be changed to `We define the following:`

spec/index.html
```diff
@@ -9946,7 +9954,7 @@ <h3>SPARQL Algebra</h3>
         <div class="defn">
           <p><b>Definition: <span id="defn_algLeftJoin">LeftJoin</span></b></p>
           <p>Let <var>Ω<sub>1</sub></var> and <var>Ω<sub>2</sub></var> be multisets of solution mappings and <var>expr</var> be an
-            expression. We define:</p>
+            <a href="#expressions">expression</a>. We define:</p>
           <p><a href="#defn_algLeftJoin" class="algFct">LeftJoin</a>(<var>Ω<sub>1</sub></var>, <var>Ω<sub>2</sub></var>, <var>expr</var>) = <a href="#defn_algFilter" class="algFct">Filter</a>(<var>expr</var>, <a href="#defn_algJoin" class="algFct">Join</a>(<var>Ω<sub>1</sub></var>,
             <var>Ω<sub>2</sub></var>)) ∪ <a href="#defn_algDiff" class="algFct">Diff</a>(<var>Ω<sub>1</sub></var>, <var>Ω<sub>2</sub></var>, <var>expr</var>)</p>
           <p><a href="#defn_Multiplicity">multiplicity</a>( <var>μ</var> | <a href="#defn_algLeftJoin" class="algFct">LeftJoin</a>(<var>Ω<sub>1</sub></var>, <var>Ω<sub>2</sub></var>, <var>expr</var>) ) =
```
TallTed, line 9960: As with above, and extending down to line 9963.

TallTed: Any other instances of `We define:</p>` and subsequent `<p>` should get similar handling.

spec/index.html
hartig: > Any other instances of `We define:</p>` and subsequent `<p>` should get similar handling.

I agree with the `We define the following:`. 

Your other suggestion (to present each block of formulas as a list) is a matter of taste, I guess.

In any case, these changes are not related to the purpose of this PR and, thus, I would keep them for a separate PR (to be created after this one is merged). I created an issue to remember this TODO: #258

Tpt (approved):

kasei (approved):

PR #257, opened 2025-08-13
Correlated EXISTS

afs:


spec/index.html
```diff
@@ -8849,16 +8849,27 @@ <h4>Variable Scope</h4>
                 <td><code>VALUES varlist { values }</code></td>
                 <td><code>v</code> is in-scope if <code>v</code> is in <code>varlist</code></td>
               </tr>
+              <tr>
+                <td>`EXISTS` and `NOT EXISTS` filters</td>
```
TallTed, line 8853:
```suggestion
                <td>`FILTERs` using `EXISTS` and `NOT EXISTS`</td>
```

spec/index.html
```diff
@@ -8849,16 +8849,27 @@ <h4>Variable Scope</h4>
...
             </tbody>
           </table>
-          <p>The variable <code>v</code> must not be in-scope at the point of the <code>(expr AS
-              v)</code> form. The scoping for <code>(expr AS v)</code> applies immediately in
-            <code>SELECT</code> expressions.</p>
+          <p>The variable <code>v</code> must not be in-scope at the point of the 
+            <code>(expr AS v)</code> form. The scoping for <code>(expr AS v)</code>
+            applies immediately in <code>SELECT</code> expressions.
+          </p>
           <p>In <code>BIND (expr AS v)</code> requires that the variable <code>v</code> is not
```
TallTed, line 8866:
```suggestion
          <p><code>BIND (expr AS v)</code> requires that the variable <code>v</code> is not
```

spec/index.html
```diff
@@ -8849,16 +8849,27 @@ <h4>Variable Scope</h4>
...
-            in-scope from the preceeding elements in the group graph pattern in which it is used.</p>
+            in-scope from the preceeding elements in the group graph pattern in which it is used.
+          </p>
           <p>In <code>SELECT</code>, the variable <code>v</code> must not be in-scope in the graph
             pattern of the <code>SELECT</code> clause, nor used in another select expression earlier in
```
TallTed, line 8870: This appears to be non-normative text. I am therefore writing to replace the lowercase RFC2119 language.
```suggestion
          <p>In <code>SELECT</code>, the variable <code>v</code> cannot be in-scope in the graph
            pattern of the <code>SELECT</code> clause, nor have been used in another select expression earlier in
```

spec/index.html
```diff
@@ -8941,13 +8953,120 @@ <h5>Collect <code>FILTER</code> Elements</h5>
...
+                    where an empty basic graph pattern start any 
+                    <a href="#rGroupGraphPattern">GroupGraphPattern</a>.
+                    It happens before the <a href="#sparqlSimplification">simplification step</a>.
+                  </div>
+                </div>
+            
+                <div class="ednote" id="note-BindingInScope">
+                  <p>
+                    One way to provide `BindingInScope` is to change `eval` to also have the current row
+                    as an argument or a "null" token. . Normally this the "null" token. This would be set in
```
TallTed, line 9002:
```suggestion
                    as an argument or a "null" token, and is normally the "null" token. This would be set in
```

spec/index.html
```diff
@@ -8941,13 +8953,120 @@ <h5>Collect <code>FILTER</code> Elements</h5>
...
+                    eval-filter.
+                  </p>
+                  <p>
+                    Another way is to have a global (to the execution) variable. In eval-exists, the
+                    old value is recorded, the global set to the new value, and reset on exit - this forms a
+                    stack for nested EXISTS.
```
TallTed, line 9008: It's not clear what is `reset on exit`. I think, maybe, replace that phrase with `the global reset to the old value on exit`, as suggested below. The hyphen should become either a full stop (as suggested below) or an em-dash (which I think doesn't work as well, here).
```suggestion
                    Another way is to have a global (to the execution) variable. In eval-exists, the
                    old value is recorded, the global set to the new value, and the global reset to the old value on exit. This forms a
                    stack for nested `EXISTS`.
```

spec/index.html
```diff
@@ -8941,13 +8953,120 @@ <h5>Collect <code>FILTER</code> Elements</h5>
...
+                  </p>
+                  <p>
+                    Modifing `eval` is better prepartion for other correlted operations.<br/>
+                    A current-row stack in EXISTS isless intrusive.
```
TallTed, line 9012:
```suggestion
                    Modifing `eval` is better preparation for other correlated operations.<br/>
                    A current-row stack in `EXISTS` is less intrusive.
```

spec/index.html
```diff
@@ -8941,13 +8953,120 @@ <h5>Collect <code>FILTER</code> Elements</h5>
...
+                  </p>
+                </div>
+                
+                <div class="defn">
+                  <b>Definition: <span id="defn_projmap">Projection Expression Variable Remapping</span></b>
+                  <p>
+                    For a projection algebra operation #sparqlProjection `Project(A, PV)` acting on algreg express `A` and with set of variables `PV`, define
```
TallTed, line 9019: My suggested `algebra regular expression` might should be `algebraic regular expression`.
```suggestion
                    For a projection algebra operation #sparqlProjection `Project(A, PV)` acting on algebra regular expression `A` with set of variables `PV`, define
```

spec/index.html
```diff
@@ -8941,13 +8953,120 @@ <h5>Collect <code>FILTER</code> Elements</h5>
...
+                  <p>
+                    For any algebra expression `X`, define the Variable Remapping `PrjMap(X)` 
+                    of algebra expression `X`:
+                  </p>
+                  <pre>PrjMap(X) = replace all project operations Project(P, PV) 
+      with ProjectMap(P, PV) for each projection in X.</pre>
+                </div>
+                <p>
+                  The outcome of `PrjMap` is independent of the order of replacement
+                  (e.g. bottom-up or top-down).
```
TallTed, line 9050:
```suggestion
                  (e.g., bottom-up or top-down).
```

TallTed (changes requested):
Incomplete, but better than no review... Some clear fixes; some less clear questions.

hartig:
@afs do you already want feedback on this one? (I am asking because it still has WiP status in the GitHub UI and you didn't say anything in the description.)

afs:
@hartig  - I was going to bring it up at the next SPARQL TF.

I use marking as "draft" to indicate its still changing so it's more like "FYI" in this state and not a request for feedback but feedback on the design is welcome. (here -  the areas are marked `@@`.)

On this PR, one area that might be of interest is whether "eval(pattern, D, G)" could be useful generalized to take a correlation 
row as an extra argument.  While it may not be necessary for EXISTS and handling the correlation row could be EXIST specific. But being mindful of future possibilities and the desire to not have too many algorithm variations, it might be a useful way to go while the spec is in WG (i.e pre features+maintenance). It isn't worse than EXIST-specific handling. The timing with regard to your current work might be quite good.

spec/index.html
TallTed: > since `FILTERs is not a keyword in SPARQL

Is your thinking that `FILTERs` (plural) is not a SPARQL keyword, wile `FILTER` (singular) is? Yes, you're correct. However, codefencing the singular and trailing it with a non-codefenced `s` to produce the plural, "`FILTER`s" generally looks awful, and often (if not always), a search on `filters` will not hit on "`FILTER`s", which I find terrible as a reader.

That said, I think your "`FILTER` statements" is better. I might suggest "`FILTER` clauses", since these do not stand on their own.

(I note that "filter" and "`FILTER`" were treated inconsistently in [SPARQL 1.1 Query Language](https://www.w3.org/TR/sparql11-query/#scopeFilters). I think "filter" should be avoided as a noun, and should generally be replaced by "`FILTER` clauses", always including the codefence around `FILTER`)

-----

> Shouldn't `BIND` with `EXISTS` or `NOT EXISTS` be considered as well?

Quite possibly. At first blush, I would guess that the answer is "yes".

-----

>  I wonder why `EXISTS` and `NOT EXISTS` need to be mentioned at all here

I will be interested in the result of this question. It might deserve its own issue.

spec/index.html
TallTed: or
```suggestion
                    as an argument or a "null" token, and is normally the latter. This would be set in
```

spec/index.html
TallTed: or, because `bottom-up` is often misunderstood
```suggestion
                  (e.g., inside-out, or outside-in).
```

TallTed: or even
```suggestion
                  (e.g., bottom-up, top-down, inside-out, or outside-in).
```

spec/index.html
```diff
@@ -10994,52 +11125,95 @@ <h3>Grammar</h3>
           section 6 <a data-cite="xml11#sec-notation">Notation</a>.</p>
         <p>Notes:</p>
         <ol>
```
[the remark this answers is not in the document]
TallTed: OK. An anticipated reader-need to refer to these notes by number is almost as compelling as a reference from somewhere in the documents we're publishing. It's certainly sufficient. I would mark this comment resolved, if I had such a button.

spec/index.html
hartig: > > since `FILTERs is not a keyword in SPARQL
> 
> Is your thinking that `FILTERs` (plural) is not a SPARQL keyword, wile `FILTER` (singular) is? Yes, you're correct.

Indeed, that's my thinking.

> However, codefencing the singular and trailing it with a non-codefenced `s` to produce the plural, "`FILTER`s" generally looks awful, and often (if not always), a search on `filters` will not hit on "`FILTER`s", which I find terrible as a reader.

I totally agree.

> That said, I think your "`FILTER` statements" is better. I might suggest "`FILTER` clauses" or "`FILTER` expressions", since these do not stand on their own.

These would be fine for me as well.

> (I note that "filter" and "`FILTER`" were treated inconsistently in [SPARQL 1.1 Query Language](https://www.w3.org/TR/sparql11-query/#scopeFilters). I think "filter" should be avoided as a noun, and should generally be replaced by "`FILTER` clauses", always including the codefence around `FILTER`.)

Yes! But that is something to be implemented in a separate PR.

TallTed: Now in https://github.com/w3c/sparql-query/issues/273

spec/index.html
```diff
@@ -10971,52 +11100,95 @@ <h3>Grammar</h3>
...
+          </li>
+          <li>
+            When tokenizing the input and choosing grammar rules, the longest match is chosen.
+          </li>
+          <li>
+            The SPARQL grammar is LL(1) when the rules with uppercased names are used as terminals.
+          </li>
+          <li>
+            There are two entry points into the grammar: <code>QueryUnit</code> for the SPARQL query language
+            and <code>UpdateUnit</code> for the SPARQL update language.
```
TallTed, line 11121:
```suggestion
            There are two entry points into the grammar: <code>QueryUnit</code> for the SPARQL Query language
            and <code>UpdateUnit</code> for the SPARQL Update language.
```

PR #259, opened 2025-08-14
Minor edit suggestions by @TallTed

hartig:
This PR implements a few edit suggestions that @TallTed had on commit https://github.com/w3c/sparql-query/commit/f4977cecbcc42fbfb28ea77a728aaddaffe6a70c in PR #245 after the PR was already merged.

@TallTed, I included all your suggestions except for [the one about code-fencing the keyword SELECT](https://github.com/w3c/sparql-query/commit/f4977cecbcc42fbfb28ea77a728aaddaffe6a70c#r163944096). That one I left out because, in this part of the spec, there are a lot of mentions of SPARQL keyword and none of them is code-fenced. Thus, code-fencing just this one mention of "SELECT" would be inconsistent with respect to the rest of this part of the spec.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/259.html" title="Last updated on Aug 15, 2025, 12:28 PM UTC (7f23944)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/259/a81b291...7f23944.html" title="Last updated on Aug 15, 2025, 12:28 PM UTC (7f23944)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -8619,7 +8619,7 @@ WHERE {
         <p>This section defines the process of converting graph patterns and solution modifiers in a
           SPARQL query string into an <a href="#defn_AlgebraicQueryExpression">algebraic
           query expression</a>. The process described converts one
-          level of query nesting, as formed by subqueries using the nested <code>SELECT</code> syntax and
+          level of query nesting as formed by subqueries using the nested <code>SELECT</code> syntax, and
           is applied recursively on subqueries. Each level consists of graph pattern matching and
           filtering, followed by the application of solution modifiers.</p>
         <p>The SPARQL query string is parsed and the abbreviations for IRIs and triple patterns given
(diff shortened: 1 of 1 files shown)


afs (approved):
Rebasing should mean this PR only has changes for spec/index.

hartig:
> > @TallTed, I included all your suggestions except for [the one about code-fencing the keyword SELECT](https://github.com/w3c/sparql-query/commit/f4977cecbcc42fbfb28ea77a728aaddaffe6a70c#r163944096). That one I left out because, in this part of the spec, there are a lot of mentions of SPARQL keyword and none of them is code-fenced. Thus, code-fencing just this one mention of "SELECT" would be inconsistent with respect to the rest of this part of the spec.
> 
> I understand your thought process here, and I generally agree with consistency. However, I believe that all SPARQL keywords in all sections should be presented as `code`, just as you can see in lines 8622 and 8718 of this PR. I think the error is that "this part of the spec" does not already do so.

I agree. But, again, if we implement such a change, it should be done in a dedicated PR.

> (I also think that the `<b>` and some `<code>` seen in the document through this PR are incorrect. I'm fairly certain that the `variable <code>v</code>` should become `variable <var>v</var>`.

Yes, and then also in the table below that paragraph, as well as in the paragraphs below the table.

> I don't know what the best marksup would be for `Let <b>P</b>, <b>P1</b>, and <b>P2</b> be graph patterns` nor `and <b>E</b>, <b>E1</b>,..., through <b>En</b> be expressions`, but I'm quite certain it's not `<b>`.)

These should all be `<var>..</var>`. Another issue here is that the table doesn't even use all these variables; it uses `expr` instead.

TallTed:
> I agree. But, again, if we implement such a change, it should be done in a dedicated PR.

Fair enough. Created #260 to move that way.

rubensworks (approved):

hartig:
@TallTed 

> > I agree. But, again, if we implement such a change, it should be done in a dedicated PR.
> 
> Fair enough. Created #260 to move that way.

Thanks for creating that other issue! Are you okay with me merging this PR now?

PR #288, opened 2025-09-22
Renames 'eval' function introduced in ALP to 'reachableTerms'

hartig:
As discussed in https://github.com/w3c/sparql-query/pull/268#discussion_r2366333242, the [definition of the ALP function](https://www.w3.org/TR/sparql12-query/#defn_evalALP_1) (used for defining property paths) introduces a function called "eval". Due to its name, this function may easily be confused with the [eval](https://www.w3.org/TR/sparql12-query/#defn_eval) function used to define the evaluation semantics of algebraic query expressions.

To avoid such confusion, this PR renames the function to _reachableTerms_ (which also reflects more closely what the function actually does).

Additionally, the PR makes this function name linkable (with an HTML anchor) and turns the two uses of the function into links that link back to the definition of the function.


Tpt (approved):

rubensworks (approved):
Looks good 👍 
Could you mention this rename in the changelog at the bottom of the spec?

kasei (approved):

PR #289, opened 2025-09-25
Add missing comparison operators for triple terms

rubensworks:
While https://github.com/w3c/sparql-query/pull/194 introduced sameValue and sameTerm support for triple terms, but `<`, `<=`, `>`, and `>=` were missing, so this PR adds them.
It also adds ORDER BY support for triple terms.

The same approach as the [RDF-star spec](https://w3c.github.io/rdf-star/cg-spec/editors_draft.html#rdf-star-operator-mapping) was followed here, with some small tweaks.

Thanks to the keen eye of @jitsedesmet for noticing this was missing!

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/289.html" title="Last updated on Dec 5, 2025, 1:54 PM UTC (2ccda34)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/289/77a17f6...2ccda34.html" title="Last updated on Dec 5, 2025, 1:54 PM UTC (2ccda34)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -12695,7 +12695,7 @@ _:x rdf:type xsd:decimal .
                 <li>Update references to XPath from 2.0 to 3.1</li>
                 <li>Define `EBV` as a functional form.</li>
                 <li>Forbid duplicated variables in `VALUES`.</li>
-                <li>Add in-between ORDER BY support for triple terms in <a href="#modOrderBy" class="sectionRef"></a>.</li>
+                <li>Add in-between term type ORDER BY support for triple terms in <a href="#modOrderBy" class="sectionRef"></a>.</li>
             </ul>
         </li>
         <li>


hartig (commented):
In addition to the suggestions in my review, I propose to remove the sentence about triple terms that is after the operator mapping table in Sec.17.3.

niklasl:
Chiming in from the side: All I can say about `>` (and `<`) is that it is sometimes very useful also for "directed difference", to avoid duplicate result rows of `?a, ?b`, `?b, ?a` from an `?a != ?b` filter, where `?a > ?b` only yields one row. Granted, `STR(?A) > STR(?b)` is a sufficient workaround (though with possible performance penalty).

jitsedesmet:
I also do not have a strong opinion, but I also think that adding support for iri comparison makes sense.
Adding support for blank nodes does not make sense to me - since it would likely depend on the labels which are often auto generated by the engines themself all with their own label generating scheme. I therefore think adding support would only fracture the landscape (imo).

If support for IRIs would be added, I think adding support for triple terms also makes sense. But like the majority here I also think adding, triple term comparison without defining IRI comparison is really confusing (to reason on what an engine would return).

spec/index.html
```diff
@@ -3674,7 +3674,7 @@ <h3>ORDER BY</h3>
         <p>The <a href="#op_lt">"&lt;" operator</a> (see the <a href="#OperatorMapping">Operator
             Mapping</a> and <a href="#operatorExtensibility">17.3.1 Operator Extensibility</a>) defines
           the relative order of pairs of <code>numerics</code>,
-          <code>xsd:strings</code>, <code>xsd:booleans</code> and <code>xsd:dateTimes</code>. Pairs of
+          <code>xsd:strings</code>, <code>xsd:booleans</code>, <code>xsd:dateTimes</code>, and <a data-cite="RDF12-CONCEPTS#dfn-triple-term">Triple Terms</a>. Pairs of
```
TallTed, line 3677: Making source more readable. A little manual `tidy`.
```suggestion
          <code>xsd:strings</code>, <code>xsd:booleans</code>, <code>xsd:dateTimes</code>, and
          <a data-cite="RDF12-CONCEPTS#dfn-triple-term">Triple Terms</a>. Pairs of
```

spec/index.html
```diff
@@ -5445,9 +5444,7 @@ <h3>Operator Mapping</h3>
               <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-triple-term">Triple Term</a></span></td>
               <td><span class="type RDFterm"><a data-cite="RDF12-CONCEPTS#dfn-triple-term">Triple Term</a></span></td>
               <td>
-                ( A.subject != B.subject ) <a href="#logical-or-operator" class="SPARQLoperator">||</a><br/>
-                ( A.predicate != B.predicate ) <a href="#logical-or-operator" class="SPARQLoperator">||</a><br/>
-                ( A.object != B.object )
+                <a data-cite="XPATH-FUNCTIONS-31#func-not">fn:not</a>(<a data-cite="XPATH-FUNCTIONS-31#func-numeric-equal">op:numeric-equal</a>(<a href="#func-compareTripleTerm">compareTripleTerm</a>(A, B), 0))
```
TallTed, line 5447: Making source more readable. A little manual `tidy`.
```suggestion
                <a data-cite="XPATH-FUNCTIONS-31#func-not">fn:not</a>(<a
                data-cite="XPATH-FUNCTIONS-31#func-numeric-equal">op:numeric-equal</a>(<a
                href="#func-compareTripleTerm">compareTripleTerm</a>(A, B), 0))
```

rubensworks:
As discussed during the last SPARQL task force meeting, this PR has been updated to include comparison for IRIs and blank nodes.
These are only defined within the scope of `compareTripleTerm` (an alternative would be to include dedicated entries for them in the operator mapping table, but that felt a bit too strong to me).
How they are implemented is up to the system.

spec/index.html
```diff
@@ -3674,7 +3674,8 @@ <h3>ORDER BY</h3>
         <p>The <a href="#op_lt">"&lt;" operator</a> (see the <a href="#OperatorMapping">Operator
             Mapping</a> and <a href="#operatorExtensibility">17.3.1 Operator Extensibility</a>) defines
           the relative order of pairs of <code>numerics</code>,
-          <code>xsd:strings</code>, <code>xsd:booleans</code> and <code>xsd:dateTimes</code>. Pairs of
+          <code>xsd:strings</code>, <code>xsd:booleans</code>, <code>xsd:dateTimes</code>, and
+          <a data-cite="RDF12-CONCEPTS#dfn-triple-term">Triple Terms</a>. Pairs of
           IRIs are ordered by comparing them as literals with datatype <code>xsd:string</code>.</p>
```
TallTed: I think we need to say something about case-_sensitive_ ordering being the norm in some environments, and case-_insensitive_ ordering in others, as well as how to override both norms, assuming that is possible anywhere if not everywhere. I know we've said in some place(s) that lexical ordering is to be code point order, which is effectively case-sensitive, but I don't think this will suffice for all use cases. So far as I knowI believe code point order will also prove problematic when trying to compare, for instance, `ä`, as a single character and as a composite character (i.e., a combination of `a` and `¨`), and in making *all the `a`s (uppercase, lowercase, with diacriticals, etc.) appear as a group and be in a predictable order within that group*.

spec/index.html
```diff
@@ -6084,6 +6116,89 @@ <h5>sameValue</h5>
...
+                An error is thrown if comparing two components results in an error.</p>
+
+            <p>Triple term component comparison
+                requires a relative ordering for pairs of IRIs and for pairs of blank nodes,
+                even if no operator extension exists for them
+                in <a href="#operatorExtensibility" class="sectionRef"></a>.
+                How this relative ordering is defined, is up to the implementation,
+                as long as it is consistent across query executions.</p>
+
+            <p class="note">For example, the relative ordering of IRIs could behave similar as in
```
```suggestion
            <p class="note">For example, the relative ordering of IRIs could behave similarly to
```

spec/index.html
```diff
@@ -6084,6 +6116,89 @@ <h5>sameValue</h5>
...
+                An error is thrown if comparing two components results in an error.</p>
+
+            <p>Triple term component comparison
+                requires a relative ordering for pairs of IRIs and for pairs of blank nodes,
+                even if no operator extension exists for them
+                in <a href="#operatorExtensibility" class="sectionRef"></a>.
+                The definition of this relative ordering is up to the implementation,
+                as long as it is consistent across query executions.</p>
+
+            <p class="note">For example, the relative ordering of IRIs could behave similarly to
```
TallTed: Sorry, I didn't read it over enough times. This should do it.
```suggestion
            <p class="note">For example, the relative ordering of IRIs here could behave similarly to that seen with
```

jitsedesmet:
@rubensworks 
Thank you for the update on the conclusion made by the working-group.

> These are only defined within the scope of compareTripleTerm (an alternative would be to include dedicated entries for them in the operator mapping table, but that felt a bit too strong to me).

I think this makes sense from a spec perspective, but I wonder what this means for query engines that have implemented comparators for bnode and iri.
So imagine my engine has a comparator for these term types, being compared like A.
Now the spec will say that within triple terms we should compare using comparator B.
That means that my engine would have two, possibly conflicting orderings. Meaning the ordering between

```
SELECT * {
    ?s ?p ?o .
    FILTER( isTriple(?o) ).
} ORDER BY (?o)
````

Would could return a different order to:

```
SELECT * {
    ?s ?p ?o .
    FILTER (isTriple(?o)) .
    BIND ( subject(?o) as ?subj ) .
} ORDER BY (?subj, ?o)
```

My engine is now not consistent with itself. Not saying this is necceserilly wrong, but it is something to be aware of.

**Comment by afs** [comment]
> These are only defined within the scope of `compareTripleTerm` (an alternative would be to include dedicated entries for them in the operator mapping table, but that felt a bit too strong to me).

I agree this is not for the operator mapping table - otherwise it would allow the use in any expression, not limited to `ORDER BY`.

It is probably worth introducing an piece of terminology for this.  Having a name means we can have a (short) section about it.
It is sort-of there, not by name, in "[15.1 ORDER BY](https://www.w3.org/TR/sparql12-query/#modOrderBy). 

"pairwise comparsion" vs "ordering comparison" -- feels a bit unnatural and clunky to repeat it over-and-over again.

"comparison" vs "collation" -- [`collation`](https://en.wikipedia.org/wiki/Collation) seems right although the dominant usage of "collation" is the natural language case.

"order collation"?

**Comment by rubensworks** [comment]
@afs Just to make sure I understand your comment correctly:

> I agree this is not for the operator mapping table - otherwise it would allow the use in any expression, not limited to ORDER BY.
> It is probably worth introducing a piece of terminology for this. Having a name means we can have a (short) section about it, define the function, and pass the function into the order processs.
> (It is sort-of there, not by name, in "[15.1 ORDER BY](https://www.w3.org/TR/sparql12-query/#modOrderBy).)

[15.1 ORDER BY](https://www.w3.org/TR/sparql12-query/#modOrderBy) explicitly refers to the "<" operator, which is extended towards triple terms within this PR.
What additional changes do you foresee exactly in ORDER BY?

> "pairwise comparsion" vs "ordering comparison" -- feels a bit unnatural and clunky to repeat it over-and-over again.
> "comparison" vs "collation" -- [collation](https://en.wikipedia.org/wiki/Collation) seems right although the dominant usage of "collation" is the natural language case.
> "order collation"?

Are you referring to the ORDER BY section here, or the new section for `compareTripleTerm`?

File: spec/index.html [hunk 6]
```diff
@@ -3684,8 +3684,10 @@ <h3>ORDER BY</h3>
           <li>Blank nodes</li>
           <li>IRIs</li>
           <li>RDF literals</li>
+          <li>Triple terms</li>
         </ol>
-        <p>SPARQL does not define a total ordering of all possible RDF terms. Here are a few examples
+        <p>SPARQL does not define a total ordering of all possible RDF terms. Implementations may
+          define total ordering through <a href="#operatorExtensibility">17.3.1 Operator Extensibility</a>. Here are a few examples
```
hartig: Section numbers shouldn't be hard coded in the text as they may change. Either simply remove the number here:

```suggestion
          define total ordering through <a href="#operatorExtensibility">operator extensibility</a>. Here are a few examples
```
... or there is a ReSpec macro that you can use (I don't have it handy right now but you can find it being used in other parts of the doc).

spec/index.html
```diff
@@ -12692,6 +12695,7 @@ <h2>Changes between SPARQL 1.1 Query Language and SPARQL 1.2 Query Language</h2>
                 <li>Update references to XPath from 2.0 to 3.1</li>
                 <li>Define `EBV` as a functional form.</li>
                 <li>Forbid duplicated variables in `VALUES`.</li>
+                <li>Add in-between ORDER BY support for triple terms in <a href="#modOrderBy" class="sectionRef"></a>.</li>
```
rubensworks:
```suggestion
                <li>Add in-between term type ORDER BY support for triple terms in <a href="#modOrderBy" class="sectionRef"></a>.</li>
```

PR #303, opened 2025-10-30
More accurate definition of the Slice algebra operator

### LINKED ISSUE #248: Translation of OFFSET and LIMIT to Slice(..)  is formally incorrect
While working on PR #245, I noticed two problems with the translation of OFFSET and LIMIT in Section [18.3.5.5 OFFSET and LIMIT](https://www.w3.org/TR/sparql12-query/#sparqlOffsetLimit).

1. For cases in which there is no LIMIT clause, the definition says: "length defaults to (size(M)-start)." While it is not explicitly stated what "size(M)" is supposed to be, I guess it is meant to refer to the number of solution mappings in M, where M is assumed to be a sequence of solution mappings. The problem is that M is not actually such a sequence. Instead, it is simply an expression of the algebraic syntax, as produced by the previous steps of the translation algorithm.
2. The Slice expression is created directly with start and length values taken from OFFSET and LIMIT, without checking these values. Then,  the [evaluation of the Slice expression](https://www.w3.org/TR/sparql12-query/#defn_evalSlice) passes these values directly to the [Slice algebra operator](https://www.w3.org/TR/sparql12-query/#defn_algSlice), which uses them without checking that they are within the bounds of the current solution sequence. What is supposed to happen if they are not?

hartig:
... as is used for OFFSET and LIMIT.

This PR closes #248

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/303.html" title="Last updated on Nov 12, 2025, 12:28 PM UTC (142053f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/303/78a901a...142053f.html" title="Last updated on Nov 12, 2025, 12:28 PM UTC (142053f)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -10200,7 +10200,7 @@ and <var>G</var> the <a href="#defn_ActiveGraph">active graph</a>.
             &nbsp; if 0 &leq; |offset| &lt; <a href="#defn_Card">Card</a>(<var>Ψ</var>)</p>
           <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, |offset|, |limit|)
             = <var>Ψ<sub>0</sub></var>
-            &nbsp; if |offset| &gt; <a href="#defn_Card">Card</a>(<var>Ψ</var>)
+            &nbsp; if |offset| &geq; <a href="#defn_Card">Card</a>(<var>Ψ</var>)
             or |limit| = 0</p>
           <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, |offset|, |limit|)
             = subseq(<var>Ψ</var>, |offset|+1, <a href="#defn_Card">Card</a>(<var>Ψ</var>))


Tpt (approved):

spec/index.html
```diff
@@ -10187,8 +10191,39 @@ <h3>SPARQL Algebra</h3>
         <p>The <a href="#defn_algReduced" class="algFct">Reduced</a> solution sequence modifier does not guarantee a defined multiplicity.</p>
         <div class="defn">
           <p><b>Definition: <span id="defn_algSlice">Slice</span></b></p>
-          <p>Let <var>Ψ</var> be a sequence of solution mappings. We define:</p>
-          <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, <var>start</var>, <var>length</var>)[<var>i</var>] = <var>Ψ</var>[<var>start</var>+<var>i</var>] for <var>i</var> = 0 to (<var>length</var>-1)</p>
+          <p>Let <var>Ψ</var> be a sequence of solution mappings, and |offset| and |length| are integers. We define:</p>
```
```suggestion
          <p>Let <var>Ψ</var> be a sequence of solution mappings, and |offset| and |length| be integers. We define:</p>
```

spec/index.html
```diff
@@ -10187,8 +10191,39 @@ <h3>SPARQL Algebra</h3>
...
+          <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, |offset|, |length|)
+            = <var>Ψ</var>
+            &nbsp; if |offset| &leq; 0
+            and |length| &geq; <a href="#defn_Card">Card</a>(<var>Ψ</var>)</p>
+          <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, |offset|, |length|)
+            = <var>Ψ<sub>0</sub></var>
+            &nbsp; if |offset| &gt; <a href="#defn_Card">Card</a>(<var>Ψ</var>)
+            or |length| &leq; 0</p>
+          <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, |offset|, |length|)
+            = subseq(<var>Ψ</var>, |offset|+1, <a href="#defn_Card">Card</a>(<var>Ψ</var>)&minus;|offset|)
```
rubensworks: I may be missing something, but it's not clear to me why `offset` is incremented with `1` here (and the line below).
I suspect `subseq` considers sequences with zero-indexed elements, so this increment seems unnecessary to me.

rubensworks: > No, subseq is one-based

Ok, then this makes sense indeed.

Should we clarify this somewhere that `subseq` is 1-based? (or perhaps say somewhere higher up that sequences are considered 1-based, e.g. near the beginning of 18.6)
But if I'm the only one that tripped over this, then we can probably keep this as-is.

hartig:
@afs can you please also look at this one. It is relatively small and shouldn't be particularly controversial.

spec/index.html
```diff
@@ -10187,8 +10191,39 @@ <h3>SPARQL Algebra</h3>
...
+            = subseq(<var>Ψ</var>, |offset|+1, |offset|+|length|)
+            &nbsp; if 0 &lt; |offset| &lt; <a href="#defn_Card">Card</a>(<var>Ψ</var>)
+            and 0 &lt; |length| &lt; <a href="#defn_Card">Card</a>(<var>Ψ</var>)&minus;|offset|</p>
+<!--           <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, <var>start</var>, <var>length</var>)[<var>i</var>] = <var>Ψ</var>[<var>start</var>+<var>i</var>] for <var>i</var> = 0 to (<var>length</var>-1)</p> -->
+          <p>where <var>Ψ<sub>0</sub></var> is the empty sequence of solution mappings
+            and, for every two integers |i| and |j|,
+            subseq(<var>Ψ</var>, |i|, |j|) is the subsequence of <var>Ψ</var>
+            that starts with the |i|-th element of <var>Ψ</var>
+            and ends with the |j|-th element of <var>Ψ</var>.
+            </p>
```
afs:
```suggestion
            Sequences are 1-based and subsequences are inclusive of both ends.
            </p>
```

spec/index.html
```diff
@@ -10737,8 +10772,10 @@ <h3>Evaluation Semantics</h3>
           </div>
           <div class="defn">
             <p><b>Definition: <span id="defn_evalSlice">Evaluation of Slice</span></b></p>
+            <p><a href="#defn_eval" class="evalFct">eval</a>( |D|(|G|), <a href="#defn_absSlice" class="absOp">Slice</a>(|L|, |offset|) ) = <a href="#defn_algSlice" class="algFct">Slice</a>( <a href="#defn_eval" class="evalFct">eval</a>(|D|(|G|), |L|), |offset| )
+            </p>
             <p>
-<a href="#defn_eval" class="evalFct">eval</a>( |D|(|G|), <a href="#defn_absSlice" class="absOp">Slice</a>(|L|, |start|, |length|) ) = <a href="#defn_algSlice" class="algFct">Slice</a>( <a href="#defn_eval" class="evalFct">eval</a>(|D|(|G|), |L|), |start|, |length| )
+<a href="#defn_eval" class="evalFct">eval</a>( |D|(|G|), <a href="#defn_absSlice" class="absOp">Slice</a>(|L|, |offset|, |length|) ) = <a href="#defn_algSlice" class="algFct">Slice</a>( <a href="#defn_eval" class="evalFct">eval</a>(|D|(|G|), |L|), |offset|, |length| )
```
afs: Personal preference: I find this pairing a bit confusing. It seems more consistent to have `start-length` or `offset-limit`

spec/index.html
```diff
@@ -10187,8 +10191,39 @@ <h3>SPARQL Algebra</h3>
         <p>The <a href="#defn_algReduced" class="algFct">Reduced</a> solution sequence modifier does not guarantee a defined multiplicity.</p>
         <div class="defn">
           <p><b>Definition: <span id="defn_algSlice">Slice</span></b></p>
-          <p>Let <var>Ψ</var> be a sequence of solution mappings. We define:</p>
-          <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, <var>start</var>, <var>length</var>)[<var>i</var>] = <var>Ψ</var>[<var>start</var>+<var>i</var>] for <var>i</var> = 0 to (<var>length</var>-1)</p>
+          <p>Let <var>Ψ</var> be a sequence of solution mappings, and |offset| and |length| be integers. We define:</p>
```
afs: Neither offset or limit can be negative by the grammar.

[15.5 LIMIT](https://www.w3.org/TR/sparql12-query/#modResultLimit)  "A limit may not be negative."

```
[28]      LimitClause    ::=  'LIMIT' INTEGER
[29]      OffsetClause   ::=  'OFFSET' INTEGER
[166]     INTEGER        ::=  [0-9]+
```

Given negative lengths in some programming languages mean "from the end"

(sorry - it can be confusing - but that's just the way the grammar is to deal with unary operators)

```suggestion
          <p>Let <var>Ψ</var> be a sequence of solution mappings, and |offset| and |length| be non-negative integers. We define:</p>
```

then simplify the definitions.

afs (changes requested):

spec/index.html
```diff
@@ -10187,8 +10191,39 @@ <h3>SPARQL Algebra</h3>
...
+            or |length| &leq; 0</p>
+          <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, |offset|, |length|)
+            = subseq(<var>Ψ</var>, |offset|+1, <a href="#defn_Card">Card</a>(<var>Ψ</var>)&minus;|offset|)
+            &nbsp; if 0 &lt; |offset| &lt; <a href="#defn_Card">Card</a>(<var>Ψ</var>)
+            and |length| &geq; <a href="#defn_Card">Card</a>(<var>Ψ</var>)&minus;|offset|</p>
+          <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, |offset|, |length|)
+            = subseq(<var>Ψ</var>, |offset|+1, |offset|+|length|)
+            &nbsp; if 0 &lt; |offset| &lt; <a href="#defn_Card">Card</a>(<var>Ψ</var>)
+            and 0 &lt; |length| &lt; <a href="#defn_Card">Card</a>(<var>Ψ</var>)&minus;|offset|</p>
+<!--           <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, <var>start</var>, <var>length</var>)[<var>i</var>] = <var>Ψ</var>[<var>start</var>+<var>i</var>] for <var>i</var> = 0 to (<var>length</var>-1)</p> -->
```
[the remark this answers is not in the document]
hartig: > Should this be removed?

Yes! Done now (see commit https://github.com/w3c/sparql-query/pull/303/commits/82ad6a48a03ee5651520506ea3a8ba6289112f34)

hartig:
@afs I have addressed all your comments now.

spec/index.html
```diff
@@ -10187,8 +10191,35 @@ <h3>SPARQL Algebra</h3>
...
+          <p>Let <var>Ψ</var> be a sequence of solution mappings, and |offset| and |limit| be non-negative integers. We define:</p>
+          <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, |offset|)
+            = <var>Ψ<sub>0</sub></var>
+            &nbsp; if |offset| &gt; <a href="#defn_Card">Card</a>(<var>Ψ</var>)</p>
+          <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, |offset|)
+            = subseq(<var>Ψ</var>, |offset|+1, <a href="#defn_Card">Card</a>(<var>Ψ</var>)&minus;|offset|)
+            &nbsp; if 0 &leq; |offset| &lt; <a href="#defn_Card">Card</a>(<var>Ψ</var>)</p>
+          <p><a href="#defn_algSlice" class="algFct">Slice</a>(<var>Ψ</var>, |offset|, |limit|)
+            = <var>Ψ<sub>0</sub></var>
+            &nbsp; if |offset| &gt; <a href="#defn_Card">Card</a>(<var>Ψ</var>)
```
afs:
```suggestion
            &nbsp; if |offset| &geq; <a href="#defn_Card">Card</a>(<var>Ψ</var>)
```

PR #316, opened 2025-11-20
Clarify LANGMATCHES handling of empty strings.

kasei:
Updated wording for the `LANGMATCHES` definition that attempts to clarify behavior on non-valid inputs. This makes clear that `false` should be returned on an empty-string tag or range.

It does **not** add any text about what should happen on tags or ranges that conform to the SPARQL grammar, but which do not conform to the more strict RFC grammar. That seems more fraught, since existing systems may support such tags, and may rely on matching behavior that does not fully conform to the RFC.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/316.html" title="Last updated on Nov 21, 2025, 10:30 PM UTC (b1424c0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/316/46e5ac5...b1424c0.html" title="Last updated on Nov 21, 2025, 10:30 PM UTC (b1424c0)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -7459,7 +7459,7 @@ WHERE {
               [[RFC4647]] section 3.3.1. Otherwise, the function returns `false`.
             </p>
               
-            <p>If either <code>language-tag</code> or <code>language-range</code> are empty
+            <p>If <code>language-tag</code>, <code>language-range</code>, or both are empty
               (and thus not a valid language tag or language range, respectively),
               the function returns `false`.
             </p>


spec/index.html
```diff
@@ -7453,14 +7453,18 @@ <h5>langMATCHES</h5>
...
-              non-empty <code>language-tag</code> string. Otherwise, the function returns `false`.
+            <p>Returns `true` if the <code>language-tag</code> (a <a data-cite="RDF12-CONCEPTS#dfn-language-tag">language tag</a>)
+               matches <code>language-range</code> (a basic language range per [[[RFC4647]]] [[RFC4647]] section 2.1)
+              according to the basic filtering scheme defined in
+              [[RFC4647]] section 3.3.1. Otherwise, the function returns `false`.
+            </p>
+              
+            <p>If either <code>language-tag</code> or <code>language-range</code> are empty
+              (and thus not a valid language tag or language range, respectively),
+              the function returns `false`.
```
Tpt: I would tend to prefer to return an error here. That `LANGMATCHES("", "")` returns `false` seems quite weird to me but returning an error seems more explainable because `""` is not a valid language tag. Also, with an error, the behavior is similar to e.g. `LANGMATCHES(<ex:a>, <ex:b>`

kasei: As I mentioned in the issue, I don't have strong feelings here, as long as the wording is clear. I think it's easy to see why `""` would result in an error (but have a tougher time with the non-empty error cases, since that might interfere with existing systems and data).

afs: > * I'm personally not sure this is an area where we should be encouraging extension.

Observation:
One potential extension would be Extended Language Range.

Let's finish this PR and circle back if that can be shown to be used somewhere.

rubensworks (approved):

spec/index.html
```diff
@@ -7453,14 +7453,18 @@ <h5>langMATCHES</h5>
...
-              [[RFC4647]] section 3.3.1. <code>language-range</code> is a basic language range per
-              [[[RFC4647]]] [[RFC4647]] section 2.1. A <code>language-range</code> of "*" matches any
-              non-empty <code>language-tag</code> string. Otherwise, the function returns `false`.
+            <p>Returns `true` if the <code>language-tag</code> (a <a data-cite="RDF12-CONCEPTS#dfn-language-tag">language tag</a>)
+               matches <code>language-range</code> (a basic language range per [[[RFC4647]]] [[RFC4647]] section 2.1)
+              according to the basic filtering scheme defined in
+              [[RFC4647]] section 3.3.1. Otherwise, the function returns `false`.
+            </p>
+              
+            <p>If either <code>language-tag</code> or <code>language-range</code> are empty
```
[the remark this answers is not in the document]
afs: I don't like `and/or`.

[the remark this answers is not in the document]
TallTed:
```suggestion
            <p>If <code>language-tag</code>, <code>language-range</code>, or both are empty
```

kasei:
Thanks for the feedback. I've applied several of the suggested changes. Will leave it a bit longer for any final reviews and then merge.

hartig (approved):

PR #322, opened 2025-11-30
vocab: rename `sparql:plus` into `sparql:add`

Tpt:
We already have `sparql:substract`, `sparql:multiply`... and XPath also uses "add". Having "plus" here sounds a bit weird to me.


afs:
https://github.com/w3c/data-shapes/pull/511

cc: @simonstey

afs (approved):

kasei (approved):

PR #327, opened 2025-12-12
Clarify that unaggregated variables in SELECT raise an error

### LINKED ISSUE #165: Clarify use of SAMPLE
Paragraph [11.4 Aggregate Projection Restrictions](https://www.w3.org/TR/sparql12-query/#aggregateRestrictions) states that <q>only expressions consisting of aggregates and constants may be projected, with one exception</q>, then gives an example of a disallowed expression with <q>Note that it would not be legal to project</q>.

However, just after it states that  <q>Other expressions, not using GROUP BY variables, or aggregates may have non-deterministic values projected from their groups using the SAMPLE aggregate.</q>. Is seems contradictory to me.
Moreover, the algorithm on section [18.2.4.1 Grouping and Aggregation](https://www.w3.org/TR/sparql12-query/#sparqlGroupAggregate) suggest this automated insertion of `SAMPLE`, leading to think this insertion of `SAMPLE` is mandatory (line `Replace V with Sample(V)`).

Example of affected query:
```sparql
SELECT ?s WHERE {
  VALUES (?s ?o) { (0 0) (1 0) }
} GROUP BY ?o
```

Jena, Blazegraph and Oxigraph all return an error for this query whereas Virtuoso returns results using `SAMPLE`.

I see multiple way forward:
1. State the behavior (error or `SAMPLE`) is implementation defined
2. Mandate error behavior
3. Mandate sample behavior

A proposal: take the "implementation defined" way to keep backward compatibility, but strongly nudge in favor of returning an error:
- in the algorithm, replace `Replace V with Sample(V)` by `Raise an error or Replace V with Sample(V)`
- rephrase <q>Other expressions, not using GROUP BY variables, or aggregates may have non-deterministic values projected from their groups using the SAMPLE aggregate.</q> into <q>Note that some implementation do not raise an error when not using GROUP BY variables, or aggregates in exception but project non-deterministic values from their groups using the SAMPLE aggregate.</q>

Tpt:
Makes it clear that inserting SAMPLE is something the user can do, not something SPARQL implementation should do automatically

Close #165

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/327.html" title="Last updated on Mar 18, 2026, 9:03 PM UTC (3f3e302)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/327/dddd137...3f3e302.html" title="Last updated on Mar 18, 2026, 9:03 PM UTC (3f3e302)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -9540,7 +9540,7 @@ Global <var>i</var> := 1   # Initially 1 for each query processed
 
 For each (<var>X</var> AS <var>Var</var>) in SELECT, each HAVING(<var>X</var>), and each ORDER BY <var>X</var> in <var>Q</var>
   For each unaggregated variable <var>V</var> in <var>X</var>
-      Replace <var>V</var> with SAMPLE(<var>V</var>)
+      Raise an error
       End
   For each aggregate <var>R</var>(<var>args</var> ; <var>scalarvals</var>) now in <var>X</var>
       # note: <var>scalarvals</var> may be omitted; if so, it is equivalent to the empty function
--- a/spec/index.html
+++ b/spec/index.html
@@ -2929,8 +2929,8 @@ WHERE {
           <p>Note that it would not be legal to project <code>STR(?z)</code> as this is not a simple
             variable expression. However, with <code>GROUP BY (STR(?z) AS ?strZ)</code> it would be
             possible to project <code>?strZ</code>.</p>
-          <p>Other expressions, not using <code>GROUP BY</code> variables, or aggregates may have
-            non-deterministic values projected from their groups using the <code>SAMPLE</code>
+          <p>For other expressions (not using <code>GROUP BY</code> variables or aggregates),
+            non-deterministic values may be requested from their groups by using the <code>SAMPLE</code>
             aggregate function explicitly.</p>
         </div>
       </section>
--- a/spec/index.html
+++ b/spec/index.html
@@ -2925,12 +2925,14 @@ WHERE {
     ?x :q ?z .
 } GROUP BY ?x (STR(?z))
         </pre>
-        <p>Note that it would not be legal to project <code>STR(?z)</code> as this is not a simple
-          variable expression. However, with <code>GROUP BY (STR(?z) AS ?strZ)</code> it would be
-          possible to project <code>?strZ</code>.</p>
-        <p>Other expressions, not using <code>GROUP BY</code> variables, or aggregates may have
-          non-deterministic values projected from their groups using the <code>SAMPLE</code>
-          aggregate.</p>
+        <div class="note">
+          <p>Note that it would not be legal to project <code>STR(?z)</code> as this is not a simple
+            variable expression. However, with <code>GROUP BY (STR(?z) AS ?strZ)</code> it would be
+            possible to project <code>?strZ</code>.</p>
+          <p>Other expressions, not using <code>GROUP BY</code> variables, or aggregates may have
+            non-deterministic values projected from their groups using the <code>SAMPLE</code>
+            aggregate function explicitly.</p>
+        </div>
       </section>
       <section id="aggregateExample2">
         <h3>Aggregate Example (with errors)</h3>


hartig (approved):

Tpt:
> I don't think agg08 shows this. It has grouping on the expression (?O1 + ?O2) and then tries to use that same expression in projection, so it isn't about a "simple expressions consisting of just a variable". 

Yes, it's not about single variable indeed. My bad.

> agg09, agg10, and agg11 similarly don't seem to be about using such a grouped "simple expression" in a non-simple projection.

It seems to me we are not focusing on the same cases. I am focusing on  "simple expressions consisting of just a variable" in the projection. I still think `agg09` (and `agg10` are relevant). Taking `agg09`:

```sparql
SELECT ?P (COUNT(?O) AS ?C)
WHERE { ?S ?P ?O } GROUP BY ?S
```

There are two possible outcomes:

1. We apply the section 11.4 Aggregate Projection Restrictions
<blockquote>In a query level which uses aggregates, only expressions consisting of aggregates and constants may be projected, with one exception. When <code>GROUP BY</code> is given with one or more simple expressions consisting of just a variable, those variables may be projected from the level.</blockquote>
Then this query is invalid because `?P` in the `SELECT` is not a grouped value.

2. We apply 18.3.4.1 Grouping and Aggregation algorithm and the line
<blockquote>Replace <var>V</var> with SAMPLE(<var>V</var>)</blockquote>
Then the query become equivalent to

```sparql
SELECT (SAMPLE(?P) AS ?P) (COUNT(?O) AS ?C)
WHERE { ?S ?P ?O } GROUP BY ?S
``` 
and we get results.

Hence my impression that the spec is inconsistent.

My MR is taking the first approach and changing algorithm but I don't feel strongly on it. I just think we should be consistent and make a choice here.

TallTed:
(I meant, as IEs in this WG, presuming that they were no longer with member organizations. I don't really care what their status was in the past group(s), nor would be in this group, as long as they participated!)

hartig:
First off, I agree with @TallTed that it would be better to have the original editors---or, more precisely, the original author(s) of this part of the spec---around here to help us understand what exactly the intention of such ambiguous sentences was.

Yet, as they do not seem to be here, we have to interpret the spec text as given. Doing so, I observe that we (several of the new editors) seem to agree that there are different ways to interpret the given text. Additionally, as illustrated by @tpt's quick test, vendors also came to different interpretations.

Therefore, as one of the new editors, I would like to fix the apparent ambiguity. And. without the previous author(s) of this part of the spec being here to help us interpreting their text—in combination with the fact that implementations are already diverging in terms of this aspect of the spec—I would say it is okay to tighten this aspect of the spec by picking the option that we think is most reasonable.

afs:
[`agg08`](https://github.com/w3c/rdf-tests/blob/main/sparql/sparql11/aggregates/agg08.rq) (reformatted, simplified) is a negative syntax test.

```sparql
PREFIX : <http://www.example.org>

SELECT ((?o1 + ?o2) AS ?o12)
WHERE { ?s :p ?o1; :q ?o2 }
GROUP BY (?o1 + ?o2)
ORDER BY ?o12
```

The user can get the effect by assigning the expression, which fixes the value as `GROUP` happens, before the various apsects of the SELECT clause. This is the point of [`agg08b`](https://github.com/w3c/rdf-tests/blob/main/sparql/sparql11/aggregates/agg08b.rq)  (reformatted, simplified)  -- "grouping by expression, done correctly"

```sparql
PREFIX : <http://www.example.org>
SELECT (?X AS ?o12)
WHERE { ?s :p ?o1; :q ?o2 }
GROUP BY (?o1 + ?o2 AS ?X)
ORDER BY ?o12
```


```sparql
SELECT ?x WHERE { ?x :p 123 } GROUP BY ?x
```

the query is covered by the text:

>For each variable V appearing outside of an aggregate
   Ai := [Aggregation](https://www.w3.org/TR/sparql12-query/#defn_absAggregation)(V, [Sample](https://www.w3.org/TR/sparql12-query/#defn_aggSample), {}, Grp)
   E := E append (V, aggi)
   i := i + 1
   End

Greg's example is one of three cases:

```
## Group key
SELECT (?x AS ?y) WHERE { ?x :p ?v } GROUP BY ?x
```
and also consider:
```
## Non-group key
SELECT (?v AS ?y) WHERE { ?x :p ?v } GROUP BY ?x
```
```
## New variable
SELECT (?z AS ?y) WHERE { ?x :p ?v } GROUP BY ?x
```

Do we agree that
* `(?x AS ?y)` is intended to work?
* `(?v AS ?y)` is `?v` as an expression and is intended to be a syntax error?
* `(?z AS ?y)` : choice -- see below.

In the text:

> For each (X AS Var) in SELECT, each HAVING(X), and each ORDER BY X in Q
  For each unaggregated variable V in X
      Replace V with SAMPLE(V)
      End

"unaggregated variable" is not mentioned anywhere else.

`X` isn't described, but

> `Replace V with SAMPLE(V)`
...
> For each aggregate R(args ; scalarvals) now in X

"in X" suggests an expression, but X as an arbitrary expression does not work without a further condition.

If `X` is used twice, it could be different values from SAMPLE.

> Note that the [Sample](https://www.w3.org/TR/sparql12-query/#defn_aggSample) function is not required to be deterministic for a given input. The only restriction is that the output value must be present in the input sequence.

In `(?v + ?v) AS ?w`, `(SAMPLE(?v) + SAMPLE(?v)) AS ?w`, `?w` is not necessarily `2*?v`.
And `SAMPLE(?v) < SAMPLE(?v)` is bizarre.

We can make that an erratum if it is not already covered.

Whether X is a variable used to give a group key value, or not, is the difference in direction between the current text and the PR change to an error.

One charitable reading of "unaggregated variable" is as a variable used as `GROUP BY` (scoping hides variables
in the group that are not part of grouping).

The `SELECT` clauses mixes up several steps, in particularly, the aggregation step and post-aggregation usage.
(aside: https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/pipe-syntax)

The `?v` in `(?v AS ?y)` could be understood is a new variable (scoping hides the inner `?v`) or as an experssion that is not a group-key variable and is unstable. The PR change relates to the latter and more consistent with the nature of agg08.

The `?z` in `(?z AS ?y)` could be understood is a new variable or as an expression that is using a non-group-key variable and (it happens to have a stable evaluation).

Overall, I think that we have can make it a principle that, after grouping, at a given query level, the set of variables allowed outside aggregates is only the variables used directly for group keys values (`GROUP BY ?x` , and `?E` in `GROUP BY (?v+1 AS ?E`)) i.e. in-scope, and any variables introduced left-to-right by SELECT AS.

That makes SELECT have aggregate functions, group keys variables and correctly introduced variables.

----

Summary:
* Define "unaggregated variable" as a variable used as a group key, including `AS ?var`. Rename.
* Add a new item to the grammar rules that use of variables out of scope after GROUP can not be used in expressions in `SELECT (expr AS ?var)` when grouping/aggregation is in effect.
* Maybe add new item to the grammar rules that only variables in-scope after GROUP, and introduced in the SELECT clause, can be used in SELECT outside aggregates.

kasei:
> * `(?x AS ?y)` is intended to work?
> * `(?v AS ?y)` is ?v as an expression and is intended to be a syntax error? (by 11.4)
> * `(?z AS ?y)` is ?z as an expression and is intended to be a syntax error? (by 11.4)

Agreed.

I'd like to add one more as the origin of my uncertainty on this issue:

```
## Grouping key used in a non-simple projected expression.
SELECT (?v+1 AS ?y) WHERE { ?x :p ?v } GROUP BY ?v
```

I think this should work also.

afs:
> I'd like to add one more as the origin of my uncertainty on this issue:
> 
> ```
> ## Grouping key used in a non-simple projected expression.
> SELECT (?v+1 AS ?y) WHERE { ?x :p ?v } GROUP BY ?v
> ```
> 
> I think this should work also.

Agreed.

Tpt:
> I can take the issue on if you want - I'd like to see that we-all agree the problem and direction before moving to text changes (see 'summary' - modified by https://github.com/w3c/sparql-query/pull/327#issuecomment-3670161469).

It would be amazing!

> Do we agree that:

Yes!

ktk:
Discussed today, @afs will work on it.

spec/index.html
```diff
@@ -2925,12 +2925,14 @@ <h3>Aggregate Projection Restrictions</h3>
...
-        <p>Other expressions, not using <code>GROUP BY</code> variables, or aggregates may have
-          non-deterministic values projected from their groups using the <code>SAMPLE</code>
-          aggregate.</p>
+        <div class="note">
+          <p>Note that it would not be legal to project <code>STR(?z)</code> as this is not a simple
+            variable expression. However, with <code>GROUP BY (STR(?z) AS ?strZ)</code> it would be
+            possible to project <code>?strZ</code>.</p>
+          <p>For other expressions (not using <code>GROUP BY</code> variables or aggregates),
+            non-deterministic values may be requested from their groups by using the <code>SAMPLE</code>
+            aggregate function explicitly.</p>
```
TallTed, line 2934: Assuming something like this remains...
```suggestion
            non-deterministic values may be requested from their groups by explicitly using the
            <code>SAMPLE</code> aggregate function.</p>
```

spec/index.html
```diff
@@ -9538,7 +9540,7 @@ <h5>Grouping and Aggregation</h5>
 
 For each (<var>X</var> AS <var>Var</var>) in SELECT, each HAVING(<var>X</var>), and each ORDER BY <var>X</var> in <var>Q</var>
   For each unaggregated variable <var>V</var> in <var>X</var>
-      Replace <var>V</var> with SAMPLE(<var>V</var>)
+      Raise an error
```
TallTed, line 9543: I absolutely support changing this text. Other changes may be better than `Raise an error`.

lisp:
our experience has been that client queries have expected the second variant in the list to succeed.

> 
> * `(?x AS ?y)` is intended to work?
> * `(?v AS ?y)` is ?v as an expression and is intended to be a syntax error? (by 11.4)
> * `(?z AS ?y)` is ?z as an expression and is intended to be a syntax error? (by 11.4)

that is, in order to accept

    SELECT (?v AS ?y) WHERE { ?x :p ?v } GROUP BY ?x

we have had to add a configuration option to satisfy expectations, to permit the lax definition.  
it would be good to have a working group decision on this before we more to enforce a strict definition.

PR #379, opened 2026-05-18
Restrict use of SELECT * in queries with implicit grouping.

### LINKED ISSUE #378: Consider adding syntactic restrictions on `SELECT *` on grouped queries
[§16.1.1 Projection](https://w3c.github.io/sparql-query/spec/#selectproject) says:

> Use of `SELECT *` is only permitted when the query does not have a `GROUP BY` clause

I think the language here should be expanded to cover any query that has grouping, regardless of the use of `GROUP BY`. Queries can introduce grouping in several ways, as described in [§ 18.3.4.1 Grouping and Aggregation](https://w3c.github.io/sparql-query/spec/#sparqlGroupAggregate):

> If the GROUP BY keyword is used, or there is implicit grouping due to the use of aggregates in HAVING or ORDER BY clauses, or in the projection, then grouping is performed by the Group function.

Should the restriction in 16.1.1 prevent use of `SELECT *` on queries that use aggregates in `HAVING` or `ORDER BY` clauses (acknowledging that such queries would be mostly trivial or non-sensical)? For example:

```
SELECT * WHERE {
	VALUES ?x { 1 2 3 4 }
}
HAVING (COUNT(*) > 0)
```

kasei:
Closes #378.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/379.html" title="Last updated on May 20, 2026, 5:08 PM UTC (fe9a7bb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/379/af0ed0f...fe9a7bb.html" title="Last updated on May 20, 2026, 5:08 PM UTC (fe9a7bb)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -12891,7 +12891,7 @@ _:x rdf:type xsd:decimal .
                   `LANGDIR`, `hasLANG`, hasLANGDIR, and `STRLANGDIR`</li>
                 <li>Define parser input as being an 
                   <a data-cite="RDF12-CONCEPTS#dfn-rdf-string">RDF string</a>. 
-                  Exclude Unicode surrogates from Unicode escape sequences.</li>
+                  Exclude Unicode surrogates from Unicode escape sequences</li>
                 <li>Remove concepts of plain and simple literals, in favor of explicit mentions of `xsd:string`</li>
                 <li>Migrate XML Schema references to 1.1. 
                 Note that for datatypes involving years, the year 1 BCE is represented by `0000` and not as `-0001`.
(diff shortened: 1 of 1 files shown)


kasei:
This doesn't (yet) add to the changelog. I would assume this would count as a normative change, although I think this aligns with the intent of the original 1.1 language.

hartig (approved):

rubensworks (approved):

kasei:
@hartig @rubensworks – any thoughts on the changelog? Do you agree this would be a normative change, and if so, can I assume your approval here would include listing it as such in the changelog?

hartig:
Yes. The two sections touched by this PR are normative and, for some engines, the change may result in adaptations to achieve conformance.

rubensworks:
Yes, I agree it would be good to add this to the changelog (no additional approval from me is needed).

> Do you agree this would be a normative change

Yes. At least in our implementation, this will require a (small) change.

Tpt (approved):

kasei:
Added this to the list of normative changes. Also improved consistency of the changelog list items so that they all omit a trailing dot.

Tpt:
@rubensworks sorry, bad manipulation in GitHub UI. No need to re-review this MR

afs:
> Do you agree this would be a normative change,

FWIW Jena does not consider this a normative change. 
It treats aggregate syntax as meaning a query has a GROUP BY.

afs (approved):

spec/index.html
```diff
@@ -12884,7 +12891,7 @@ <h2>Changes between SPARQL 1.1 Query Language and SPARQL 1.2 Query Language</h2>
                   `LANGDIR`, `hasLANG`, hasLANGDIR, and `STRLANGDIR`</li>
                 <li>Define parser input as being an 
                   <a data-cite="RDF12-CONCEPTS#dfn-rdf-string">RDF string</a>. 
-                  Exclude Unicode surrogates from Unicode escape sequences.</li>
+                  Exclude Unicode surrogates from Unicode escape sequences</li>
```
TallTed, line 12894: I believe this should be 2 list items, not 1. On that basis, a dot should be deleted (though I disagree with this practice when (some of) the list items are complete sentences, on which basis some other list items should be edited).
```suggestion
                <li>Define parser input as being an 
                  <a data-cite="RDF12-CONCEPTS#dfn-rdf-string">RDF string</a></li>
                <li>Exclude Unicode surrogates from Unicode escape sequences</li>
```

spec/index.html
```diff
@@ -12893,13 +12900,14 @@ <h2>Changes between SPARQL 1.1 Query Language and SPARQL 1.2 Query Language</h2>
                 for details.
```
TallTed, line 12900: This is the third complete sentence in this list item. I agree that it should end with a dot. But for consistency, *all* list items should end with dots, and for best consistency all list items should be written as complete sentences (which I believe most to be already).

kasei: @TallTed – This has already been merged. A new PR is probably the best approach for the changes you suggested here and above.

spec/index.html
TallTed: moved to issue #386 and pr #387

PR #380, opened 2026-05-22
Update Aggregate Projection Restrictions.

### LINKED ISSUE #377: Improve wording involving aggregate projection restrictions and scoping
After discussion on today's SPARQL TF call, I'm creating this issue to track changes needed to spec text relating to the use of variables in projection and project expressions in conjunction with aggregation (and aggregation-adjacent operations like grouping). Will follow up with proposed spec text changes here or in a PR.

kasei:
Since there hasn't been any discussion on #377, I'm moving the proposed changes to this PR. Copying from that issue:

> The big changes here are:
> 
> * explicitly includes any query level that uses any of the aggregation mechanisms (spelled out by [18.3.4.1 Grouping and Aggregation](https://w3c.github.io/sparql-query/spec/#sparqlGroupAggregate) as the use of `GROUP BY` or "the use of aggregates in `HAVING` or `ORDER BY` clauses, or in the projection").
> * changes the language about the restriction on expressions from only "consisting of aggregates and constants" or simple `GROUP BY` variables to what I think is more explicit language about conditions that *every* variable used in SELECT expressions in these aggregating queries MUST satisfy
> * explicitly state that if the conditions on variables in SELECT expressions are not satisfied that the query is syntactically invalid

Closes #377

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/380.html" title="Last updated on May 28, 2026, 3:41 PM UTC (ba816b8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/380/af0ed0f...ba816b8.html" title="Last updated on May 28, 2026, 3:41 PM UTC (ba816b8)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -2960,8 +2960,8 @@ HAVING(AVG(?size) &gt; 10)
       </section>
       <section id="aggregateRestrictions">
         <h3>Aggregate Projection Restrictions</h3>
-        <p>In a query level which uses grouping (either through the explicit use of aggregates in projection,
-          <code>HAVING</code>, or <code>ORDER BY</code> clauses, or by the use of a <code>GROUP BY</code> clause),
+        <p>In a query level which uses grouping (either by the explicit use of a <code>GROUP BY</code> clause
+          or through the use of aggregates in projection, <code>HAVING</code>, or <code>ORDER BY</code> clauses),
           all variables appearing in projection or SELECT expressions MUST satisfy exactly one of the following conditions:</p>
         <ul>
         <li>the variable is aggregated</li>
--- a/spec/index.html
+++ b/spec/index.html
@@ -2960,10 +2960,16 @@ HAVING(AVG(?size) &gt; 10)
       </section>
       <section id="aggregateRestrictions">
         <h3>Aggregate Projection Restrictions</h3>
-        <p>In a query level which uses aggregates, only expressions consisting of aggregates and
-          constants may be projected, with one exception. When <code>GROUP BY</code> is given with one
-          or more simple expressions consisting of just a variable, those variables may be projected
-          from the level.</p>
+        <p>In a query level which uses grouping (either through the explicit use of aggregates in projection,
+          <code>HAVING</code>, or <code>ORDER BY</code> clauses, or by the use of a <code>GROUP BY</code> clause),
+          all variables appearing in projection or SELECT expressions MUST satisfy one of the following conditions:</p>
+        <ul>
+        <li>the variable is aggregated</li>
+        <li>the variable, <var>V</var>, appears as a named <code>GROUP BY</code> variable (with the <code>GROUP BY</code> expression consisting of just the variable <var>V</var>, or having the form <code>(expr AS <var>V</var>)</code>)</li>
+        <li>the variable is introduced by an earlier SELECT expression in the same SELECT clause</li>
+        </ul>
+        <p>If such a variable does not satisfy one of these conditions, the query is syntactically invalid.</p>
+
         <p>For example, the following query is legal as ?x is given as a <code>GROUP BY</code>
           term.</p>
         <pre class="query nohighlight">
@@ -2974,12 +2980,11 @@ WHERE {
     ?x :q ?z .
 } GROUP BY ?x (STR(?z))
         </pre>
-        <p>Note that it would not be legal to project <code>STR(?z)</code> as this is not a simple
-          variable expression. However, with <code>GROUP BY (STR(?z) AS ?strZ)</code> it would be
+        <p>Note that it would not be legal to project <code>STR(?z)</code> as this expression is neither a simple variable,
+          nor a named <code>GROUP BY</code> expression. However, with <code>GROUP BY (STR(?z) AS ?strZ)</code> it would be
           possible to project <code>?strZ</code>.</p>
-        <p>Other expressions, not using <code>GROUP BY</code> variables, or aggregates may have
-          non-deterministic values projected from their groups using the <code>SAMPLE</code>
-          aggregate.</p>
+        <p>Other expressions which use variables not satisfying the above conditions may be
+          projected from their groups using the <code>SAMPLE</code> aggregate.</p>
       </section>
       <section id="aggregateExample2">
         <h3>Aggregate Example (with errors)</h3>
--- a/spec/index.html
+++ b/spec/index.html
@@ -2965,7 +2965,8 @@ HAVING(AVG(?size) &gt; 10)
           every variable that appears in projection or SELECT expressions of that query level
           MUST satisfy exactly one of the following conditions:</p>
         <ul>
-        <li>the variable is aggregated</li>
+        <li>the variable appears as part of a sub-expression within the SELECT expression and this sub-expression is used as an argument to an aggregate (e.g., <code>MIN(?v)</code> or <code>AVG(?v+2)</code>)
+        </li>
         <li>the variable, <var>V</var>, appears as a named <code>GROUP BY</code> variable (with the corresponding <code>GROUP BY</code> expression consisting of just the variable <var>V</var>, or having the form <code>(expr AS <var>V</var>)</code>)</li>
         <li>the variable is introduced by an earlier SELECT expression in the same SELECT clause</li>
         </ul>
--- a/spec/index.html
+++ b/spec/index.html
@@ -2962,7 +2962,7 @@ HAVING(AVG(?size) &gt; 10)
         <h3>Aggregate Projection Restrictions</h3>
         <p>In a query level which uses grouping (either through the explicit use of aggregates in projection,
           <code>HAVING</code>, or <code>ORDER BY</code> clauses, or by the use of a <code>GROUP BY</code> clause),
-          all variables appearing in projection or SELECT expressions MUST satisfy one of the following conditions:</p>
+          all variables appearing in projection or SELECT expressions MUST satisfy exactly one of the following conditions:</p>
         <ul>
         <li>the variable is aggregated</li>
         <li>the variable, <var>V</var>, appears as a named <code>GROUP BY</code> variable (with the <code>GROUP BY</code> expression consisting of just the variable <var>V</var>, or having the form <code>(expr AS <var>V</var>)</code>)</li>
--- a/spec/index.html
+++ b/spec/index.html
@@ -2962,7 +2962,8 @@ HAVING(AVG(?size) &gt; 10)
         <h3>Aggregate Projection Restrictions</h3>
         <p>In a query level which uses grouping (either by the explicit use of a <code>GROUP BY</code> clause
           or through the use of aggregates in projection, <code>HAVING</code>, or <code>ORDER BY</code> clauses),
-          all variables appearing in projection or SELECT expressions MUST satisfy exactly one of the following conditions:</p>
+          every variable that appears in projection or SELECT expressions of that query level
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -2960,10 +2960,16 @@ <h3>HAVING</h3>
       </section>
       <section id="aggregateRestrictions">
         <h3>Aggregate Projection Restrictions</h3>
-        <p>In a query level which uses aggregates, only expressions consisting of aggregates and
-          constants may be projected, with one exception. When <code>GROUP BY</code> is given with one
-          or more simple expressions consisting of just a variable, those variables may be projected
-          from the level.</p>
+        <p>In a query level which uses grouping (either through the explicit use of aggregates in projection,
+          <code>HAVING</code>, or <code>ORDER BY</code> clauses, or by the use of a <code>GROUP BY</code> clause),
+          all variables appearing in projection or SELECT expressions MUST satisfy one of the following conditions:</p>
```
rubensworks: AFAICS, these conditions are mutually exclusive, so we could say:
```suggestion
          all variables appearing in projection or SELECT expressions MUST satisfy exactly one of the following conditions:</p>
```

rubensworks (approved):

Tpt (approved):

spec/index.html
```diff
@@ -2960,10 +2960,16 @@ <h3>HAVING</h3>
       </section>
       <section id="aggregateRestrictions">
         <h3>Aggregate Projection Restrictions</h3>
-        <p>In a query level which uses aggregates, only expressions consisting of aggregates and
-          constants may be projected, with one exception. When <code>GROUP BY</code> is given with one
-          or more simple expressions consisting of just a variable, those variables may be projected
-          from the level.</p>
+        <p>In a query level which uses grouping (either through the explicit use of aggregates in projection,
+          <code>HAVING</code>, or <code>ORDER BY</code> clauses, or by the use of a <code>GROUP BY</code> clause),
```
hartig: I would say that an "explicit" use of grouping is via a GROUP BY clause; the other cases are implicit use of grouping.
```suggestion
        <p>In a query level which uses grouping (either by the explicit use of a <code>GROUP BY</code> clause
          or through the use of aggregates in projection, <code>HAVING</code>, or <code>ORDER BY</code> clauses),
```

spec/index.html
```diff
@@ -2960,10 +2960,16 @@ <h3>HAVING</h3>
...
+          all variables appearing in projection or SELECT expressions MUST satisfy exactly one of the following conditions:</p>
```
hartig:
```suggestion
          every variable that appears in projection or SELECT expressions of that query level
          MUST satisfy exactly one of the following conditions:</p>
```

spec/index.html
```diff
@@ -2960,10 +2960,16 @@ <h3>HAVING</h3>
...
+        <ul>
+        <li>the variable is aggregated</li>
```
hartig: What does this mean?

spec/index.html
```diff
@@ -2960,10 +2960,16 @@ <h3>HAVING</h3>
...
+        <li>the variable, <var>V</var>, appears as a named <code>GROUP BY</code> variable (with the <code>GROUP BY</code> expression consisting of just the variable <var>V</var>, or having the form <code>(expr AS <var>V</var>)</code>)</li>
```
hartig: The part in parenthesis confused me a bit, mainly because of the use of the article "the" in "the GROUP BY expression." As this sounds like there is only one "GROUP BY expression," I was assuming that the term "GROUP BY expression" is synonymous to GROUP BY clause. Based on this assumption, my initial interpretation of the text in parenthesis was that the GROUP BY clause can have either only the variable or only the `(expr AS V)'  form (but not, say, multiple variables). To avoid others coming to the same interpretation, I suggest:
```suggestion
        <li>the variable, <var>V</var>, appears as a named <code>GROUP BY</code> variable (with the corresponding <code>GROUP BY</code> expression consisting of just the variable <var>V</var>, or having the form <code>(expr AS <var>V</var>)</code>)</li>
```

spec/index.html
```diff
@@ -2974,12 +2980,11 @@ <h3>Aggregate Projection Restrictions</h3>
     ?x :q ?z .
 } GROUP BY ?x (STR(?z))
         </pre>
-        <p>Note that it would not be legal to project <code>STR(?z)</code> as this is not a simple
-          variable expression. However, with <code>GROUP BY (STR(?z) AS ?strZ)</code> it would be
+        <p>Note that it would not be legal to project <code>STR(?z)</code> as this expression is neither a simple variable,
+          nor a named <code>GROUP BY</code> expression. However, with <code>GROUP BY (STR(?z) AS ?strZ)</code> it would be
```
hartig: The following suggestion is to remain consistent with the example query above, and it may also be helpful in the context of the confusion that I mention in my other comment.
```suggestion
          nor a named <code>GROUP BY</code> expression. However, with <code>GROUP BY ?x (STR(?z) AS ?strZ)</code> it would be
```

hartig (changes requested):

spec/index.html
kasei: This is terminology already used in the spec, but perhaps we could make it more explicit with something like this:

* the variable appears as part of a sub-expression that is used as an argument to an aggregate (e.g. `MIN(?v)` or `AVG(?v+2)`)

kasei:
Thanks @hartig. Very good suggestions. I committed all but one, which I commented on directly.

spec/index.html
hartig: Yes, it would be very helpful to make that explicit.

Your proposal now may be improved further by also saying explicitly what kind of "sub-expression" it talks about (which is relevant in particular because the next bullet point also talks about expressions, but there the focus is on the GROUP BY expressions and not the SELECT expressions).
```suggestion
        <li>the variable appears as part of a sub-expression within the SELECT expression and this sub-expression is used as an argument to an aggregate (e.g., <code>MIN(?v)</code> or <code>AVG(?v+2)</code>)
        </li>
```

spec/index.html
```diff
@@ -2960,10 +2960,17 @@ <h3>HAVING</h3>
...
       <section id="aggregateRestrictions">
         <h3>Aggregate Projection Restrictions</h3>
-        <p>In a query level which uses aggregates, only expressions consisting of aggregates and
-          constants may be projected, with one exception. When <code>GROUP BY</code> is given with one
-          or more simple expressions consisting of just a variable, those variables may be projected
-          from the level.</p>
+        <p>In a query level which uses grouping (either by the explicit use of a <code>GROUP BY</code> clause
+          or through the use of aggregates in projection, <code>HAVING</code>, or <code>ORDER BY</code> clauses),
+          every variable that appears in projection or SELECT expressions of that query level
+          MUST satisfy exactly one of the following conditions:</p>
```
hartig: Now that I understand what the first bullet point is about, I would say that "exactly one" is too restrictive here. It would also be okay if a variable satisfies both the first and the second condition. Likewise, it would be okay if a variable satisfies both the second and the third condition. It may even be the case that a variable satisfies the first condition in multiple different ways (by being an argument to multiple aggregates within the corresponding SELECT clause).

Having said that, I don't think we can simply replace "exactly one" by "at least one" because, then, a query such as the following would be permitted (because ?x satisfies the first condition).
```sparql
SELECT ?x (MIN(?x) AS ?y) {
 # ... ?z ...
}
GROUP BY ?z
```
The underlying issue is that the conditions are actually not about the variables per se, but about (all) the occurrences of the variables within the SELECT clause.

kasei: Agreed it's about occurrences, not the variables themselves. I'm hesitant to get to into the weeds with this language, but I'll think about how we might tweak it to talk about occurrences.

kasei: Perhaps it's as simple as introducing "occurrence":

> every **occurrence of a** variable that appears in projection or SELECT expressions of that query level MUST satisfy exactly one of the following conditions:
…
> If such a variable **occurrence** does not satisfy one of these conditions, the query is syntactically invalid.

?

hartig: Yes! Introducing "occurrence" in this way works, together with just saying "one of" instead of saying "exactly one of."

kasei: Done.

hartig (approved):
Looks good now!

spec/index.html
```diff
@@ -2960,10 +2960,18 @@ <h3>HAVING</h3>
...
-          from the level.</p>
+        <p>In a query level which uses grouping (either by the explicit use of a <code>GROUP BY</code> clause
+          or through the use of aggregates in projection, <code>HAVING</code>, or <code>ORDER BY</code> clauses),
+          every occurrence of a variable that appears in projection or SELECT expressions of that query level
+          MUST satisfy one of the following conditions:</p>
+        <ul>
+        <li>the variable appears as part of a sub-expression within the SELECT expression and this sub-expression is used as an argument to an aggregate (e.g., <code>MIN(?v)</code> or <code>AVG(?v+2)</code>)
+        </li>
+        <li>the variable, <var>V</var>, appears as a named <code>GROUP BY</code> variable (with the corresponding <code>GROUP BY</code> expression consisting of just the variable <var>V</var>, or having the form <code>(expr AS <var>V</var>)</code>)</li>
+        <li>the variable is introduced by an earlier SELECT expression in the same SELECT clause</li>
```
afs, line 2971: Minor suggestion - approved anyway.

Put the `GROUP BY` case first. IMO It is the most immediate to understand.

```suggestion
        <li>the variable, <var>V</var>, appears as a named <code>GROUP BY</code> variable (with the corresponding <code>GROUP BY</code> expression consisting of just the variable <var>V</var>, or having the form <code>(expr AS <var>V</var>)</code>)</li>
        <li>the variable appears as part of a sub-expression within the SELECT expression and this sub-expression is used as an argument to an aggregate (e.g., <code>MIN(?v)</code> or <code>AVG(?v+2)</code>)
        </li>
        <li>the variable is introduced by an earlier SELECT expression in the same SELECT clause</li>
```

afs (approved):

spec/index.html
```diff
@@ -2960,10 +2960,18 @@ <h3>HAVING</h3>
...
       <section id="aggregateRestrictions">
         <h3>Aggregate Projection Restrictions</h3>
-        <p>In a query level which uses aggregates, only expressions consisting of aggregates and
-          constants may be projected, with one exception. When <code>GROUP BY</code> is given with one
-          or more simple expressions consisting of just a variable, those variables may be projected
-          from the level.</p>
+        <p>In a query level which uses grouping (either by the explicit use of a <code>GROUP BY</code> clause
+          or through the use of aggregates in projection, <code>HAVING</code>, or <code>ORDER BY</code> clauses),
+          every occurrence of a variable that appears in projection or SELECT expressions of that query level
+          MUST satisfy one of the following conditions:</p>
```
TallTed, line 2966:
```suggestion
        <p>In a query level which uses grouping (either by the explicit use of a <code>GROUP BY</code> clause,
          or by the use of an aggregate in projection, a <code>HAVING</code> clause, or an <code>ORDER BY</code> clause),
          every occurrence of a variable that appears in a projection or a SELECT expression of that query level
          MUST satisfy one of the following conditions:</p>
```

kasei: @TallTed – This has already been merged. A new PR is probably the best approach for this change.

TallTed: moved to issue #386 and pr #387

PR #388, opened 2026-06-11
Fix triple patterns being acceptable in subject position of triple patterns

rubensworks:
This used to be allowed in SPARQL-star, but not anymore.

We should still allow it in property paths, as we can have inverse paths.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/388.html" title="Last updated on Jun 11, 2026, 12:58 PM UTC (fea6ecf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/388/7cee218...fea6ecf.html" title="Last updated on Jun 11, 2026, 12:58 PM UTC (fea6ecf)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -8568,9 +8568,8 @@ WHERE {
               that is defined inductively as follows:
               If</p>
             <ul style="list-style-type: none;">
-              <li>|s| is an <a data-cite="RDF12-CONCEPTS#dfn-rdf-term">RDF term</a>,
-                a <a href="#defn_QueryVariable">variable</a>,
-                or a triple pattern,</li>
+              <li>|s| is an <a data-cite="RDF12-CONCEPTS#dfn-rdf-term">RDF term</a>
+                or a <a href="#defn_QueryVariable">variable</a>,</li>
               <li>|p| is an <a data-cite="RDF12-CONCEPTS#dfn-iri">IRI</a>
                 or a <a href="#defn_QueryVariable">variable</a>, and</li>
               <li>|o| is an <a data-cite="RDF12-CONCEPTS#dfn-rdf-term">RDF term</a>,


Tpt (approved):

kasei (approved):

afs:
I double checked the grammar - subjects are `VarOrTerm` which includes triple terms.

afs (approved):

hartig:
Hold on. This inverts the change I did in PR #335. There is a reason for permitting triple patterns in the subject position of other triple patterns. I even mentioned this reason in a Note added in the initial version of PR #335, but that Note [was requested to be changed](https://github.com/w3c/sparql-query/pull/335#discussion_r2662496091) during the discussion of that PR.

The reason is that, during the translation of property path patterns, a triple pattern may be created that contains another triple pattern in its subject position. This may happen through each of the first three rules in the translation table in [Section 18.3.2.4 Translate Property Path Patterns](https://www.w3.org/TR/sparql12-query/#sparqlTranslatePathPatterns), for which you need to consider that the subject and the object of a [property path pattern](https://www.w3.org/TR/sparql12-query/#defn_PropertyPathPattern) (i.e., the _x_ and the _y_ in the translation table) may be a triple pattern.

> @afs:  I double checked the grammar - subjects are `VarOrTerm` which includes triple terms.

In fact, it includes [`TripleTerm`](https://www.w3.org/TR/sparql12-query/#rTripleTerm), which permits a variable in each of the three positions. In other words, `TripleTerm` admits triple patterns, not just triples.

hartig (changes requested):
Reject for the reason mentioned in the comment that I just wrote.

afs:
> > @afs:  I double checked the grammar - subjects are `VarOrTerm` which includes triple terms.
> 
> In fact, it includes [`TripleTerm`](https://www.w3.org/TR/sparql12-query/#rTripleTerm), which permits a variable in each of the three positions. In other words, `TripleTerm` admits triple patterns, not just triples.

I'm not sure of the point here. "includes triple terms" does not mean to the exclusion of patterns.

Aside:
The non-variable case is `TripleTermData`.
Maybe `TripleTerm` could be renamed e.g. `TripleTermPattern`, with consequences of that on nearby rules (and the table width/wrappping!).  It is not the only case : `TriplesBlock` is patterns as well.

afs:
> We should still allow it in property paths, as we can have inverse paths.

The _syntax_ needs to be "symmetric RDF+variables" for practical reasons (size of grammar)


It may not be possible to rewrite it to be strict anyway because inverse can be combined with other patterns: e.g. `:p/(^:q)*`).

[Basic Graph Pattern Matching](https://www.w3.org/TR/sparql12-query/#BGPsparql) does support an intuition that variables can be substituted for RDF terms and that is then a "can't match".

rubensworks:
Thanks for your comment @hartig. I forgot about your PR, and wrongly assumed the current text was just an oversight.

Given the discussion in #335 and above, it may make sense to introduce a "Generalized Triple Pattern" definition besides the current "Triple Pattern" definition.
The property path translations could then make use of the "Generalized Triple Pattern" definition, while other parts of the spec could just rely on the stricter "Triple Pattern" definition.
This would make the definitions closer to their intended functioning.

hartig:
> > > @afs:  I double checked the grammar - subjects are `VarOrTerm` which includes triple terms.
> > 
> > 
> > In fact, it includes [`TripleTerm`](https://www.w3.org/TR/sparql12-query/#rTripleTerm), which permits a variable in each of the three positions. In other words, `TripleTerm` admits triple patterns, not just triples.
> 
> I'm not sure of the point here. "includes triple terms" does not mean to the exclusion of patterns.

That's true. I just wanted to mention it in the context of the rest of my comment, to emphasize that the grammar also permits triple patterns that have triple patterns in their subject position (rather than only triple terms).

> @rubensworks: it may make sense to introduce a "Generalized Triple Pattern" definition besides the current "Triple Pattern" definition.
> The property path translations could then make use of the "Generalized Triple Pattern" definition, while other parts of the spec could just rely on the stricter "Triple Pattern" definition.

I don't think that introducing such an extra concept is particularly useful. The issue is that such "generalized triple patterns"---after being produced as a result of the [translation of property path patterns](https://www.w3.org/TR/sparql12-query/#sparqlTranslatePathPatterns) (as discussed above)---are collected to form a BGP as per [Section 18.3.2.5 Translate Basic Graph Patterns](https://www.w3.org/TR/sparql12-query/#sparqlTranslateBasicGraphPatterns). Hence, we would need to either change the definition of BGPs to say that they are sets of such "generalized triple patterns" or introduce another new notion of a "generalized BGP". The latter may then require further changes downstream.

afs:
@rubensworks -- did this come up for you where it created a problem?

An implementation doesn't need to wait until BGP patching to "know" that a triple pattern in the subject can not match.

(I'm presuming "generalized triple patterns" is actually "symmetric RDF triple patterns" - "Generalized RDF" is too general)

rubensworks:
> I don't think that introducing such an extra concept is particularly useful.

@hartig The cost may not be worth it indeed.

> @rubensworks -- did this come up for you where it created a problem?

@afs No, there was no problem implementation-wise. I was writing a paper, for which I needed a reference for a triple pattern's definition. So I looked at the spec, and (wrongfully) assumed we forgot to update it.

Unless anyone else has some other concerns with this, or sees reasons to go forward with a dedicated symmetric/generalized triple pattern definition, I'm ok with closing this PR.

PR #399, opened 2026-06-23
Version from media type

afs:
When `version=` applies.

Text based on the text in Turtle.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/399.html" title="Last updated on Jun 25, 2026, 10:41 AM UTC (f71a525)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/399/da3537b...f71a525.html" title="Last updated on Jun 25, 2026, 10:41 AM UTC (f71a525)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1403,14 +1403,27 @@ _:b0  :p        "v" .
                   :myreifier :tripleAdded ?date .
               }
 </pre>
+        <p>
+          <p>
+            The [[[SPARQL12-PROTOCOL]]] also provides a 
+            <a data-cite="sparql12-protocol/#section-version-announcement"
+               >version announcement</a> using the `version` parameter
+            of the <a href=#mediaType>Media Type</a>.
+            This is considered if there is no `VERSION` directive.
+        </p>
+
         <section id="versionLabels">
           <h4>Version Labels</h4>
           <p>
-            A <dfn class="export">SPARQL version label</dfn> is a string that identifies the syntax and semantics conformance
-            for the SPARQL query.
+            A <dfn class="export">SPARQL version label</dfn> is a string that identifies 
+            the syntax and semantics conformance for the SPARQL query.
           </p>
           <p class="note">
-            Even though the version label strings in SPARQL are the same as <a data-cite="RDF12-CONCEPTS#defined-version-labels">the version labels defined by RDF</a>, their meaning is different. Namely, the SPARQL version labels refer to SPARQL syntax and semantics, while the RDF version labels refer to RDF syntax and semantics.
+            Even though the version label strings in SPARQL are the same
+            as <a data-cite="RDF12-CONCEPTS#defined-version-labels">the version labels defined by
+            RDF</a>, their meaning is different. Specifically, the SPARQL version labels refer to
+            SPARQL syntax and semantics, while the RDF version labels refer to RDF syntax 
+            and semantics.
           </p>
           <table id="tab-version-labels" class="simple">
             <caption>SPARQL Version Labels</caption>


spec/index.html
```diff
@@ -1395,6 +1395,13 @@ <h3>Version Announcement</h3>
                   :myreifier :tripleAdded ?date .
               }
 </pre>
+        <p>
+          [=Version labels=] can also be given by the `version` parameter of the
```
hartig: "Version labels" in this sentence links to the notion of version labels in RDF Concepts. It would be better to link to the version labels in this document.

```suggestion
          <a href="#versionLabels">Version labels</a> can also be given by the `version` parameter of the 
```

spec/index.html
```diff
@@ -1395,6 +1395,13 @@ <h3>Version Announcement</h3>
                   :myreifier :tripleAdded ?date .
               }
 </pre>
+        <p>
+          [=Version labels=] can also be given by the `version` parameter of the 
+          <a href=#mediaType>Media Type</a>. In the absence of a current
+          version directive, the version specified as part of the Media Type is
```
hartig: To be consistent with the first sentence of the section:

```suggestion
          `VERSION` directive, the version specified as part of the Media Type is
```

spec/index.html
```diff
@@ -1395,6 +1395,13 @@ <h3>Version Announcement</h3>
                   :myreifier :tripleAdded ?date .
               }
 </pre>
+        <p>
+          [=Version labels=] can also be given by the `version` parameter of the 
+          <a href=#mediaType>Media Type</a>. In the absence of a current
```
hartig: I would remove "current" because it is not clear what really it refers to. The text is fine without this word.
```suggestion
          <a href=#mediaType>Media Type</a>. In the absence of a
```

spec/index.html
```diff
@@ -1395,6 +1395,13 @@ <h3>Version Announcement</h3>
...
+          version directive, the version specified as part of the Media Type is
+          considered.
```
rubensworks: Within this text, I would consider also referring to version announcement in the SPARQL protocol (https://www.w3.org/TR/sparql12-protocol/#section-version-announcement) as an example of where this media type matters. (there are some concrete examples on content negotiation in that document)

rubensworks (changes requested):

spec/index.html
afs: good idea

spec/index.html
```diff
@@ -1403,14 +1403,26 @@ <h3>Version Announcement</h3>
...
-            A <dfn class="export">SPARQL version label</dfn> is a string that identifies the syntax and semantics conformance
-            for the SPARQL query.
+            A <dfn class="export">SPARQL version label</dfn> is a string that identifies 
+            the syntax and semantics conformance for the SPARQL query.
           </p>
           <p class="note">
-            Even though the version label strings in SPARQL are the same as <a data-cite="RDF12-CONCEPTS#defined-version-labels">the version labels defined by RDF</a>, their meaning is different. Namely, the SPARQL version labels refer to SPARQL syntax and semantics, while the RDF version labels refer to RDF syntax and semantics.
+            Even though the version label strings in SPARQL are the same
+            as <a data-cite="RDF12-CONCEPTS#defined-version-labels">the version labels defined by
+            RDF</a>, their meaning is different. Namely, the SPARQL version labels refer to SPARQL
```
```suggestion
            RDF</a>, their meanings are different. Specifically, the SPARQL version labels refer to SPARQL
```

hartig (approved):

kasei (approved):

rubensworks (approved):

spec/index.html
afs: (this is just reformatted text not new to this PR)

"their meanings" can also give a reading that each has multiple meanings.

Namely => Specifically done

TallTed: OK

spec/index.html
```diff
@@ -1403,14 +1403,27 @@ <h3>Version Announcement</h3>
                   :myreifier :tripleAdded ?date .
               }
 </pre>
+        <p>
+          <p>
+            The [[[SPARQL12-PROTOCOL]]] also provides a 
+            <a data-cite="sparql12-protocol/#section-version-announcement"
+               >version announcement</a> using the `version` parameter
+            of the <a href=#mediaType>Media Type</a>.
+            This is considered if there is no `VERSION` directive.
```
TallTed, line 1412: I think that this should say that the Media Type `version` parameter is *only* considered if there is no `VERSION` directive in the SPARQL query and/or that if there's a conflict between the Media Type `version` parameter and the VERSION` directive in the SPARQL query, the latter should govern. So either --
```suggestion
            This is only considered if there is no `VERSION` directive.
```

TallTed: -- or --
```suggestion
            This is considered if there is no `VERSION` directive.
            If there's a conflict between the Media Type `version` parameter
            and the VERSION` directive in the SPARQL query,
            the latter governs.
```

kasei: This has been merged, so probably needs a new PR, but I disagree with the suggested change here. We already have cases where if the query string and the protocol disagree, the protocol details take precedence (e.g. dataset specification). The Protocol spec [already has language](https://w3c.github.io/sparql-protocol/spec/#section-version-announcement) about this being the case for `VERSION` as well, and I think that's correct. But that probably means we should clarify the text just added in this PR to describe what happens if versions are specified in both places.

afs: "This is considered if there is no `VERSION` directive."  The media type is only looked at if there is no `VERSION`. 

The Turtle text, which was the starting point, is:

> In the absence of a current version directive, any version specified as part of the [Media Type](https://www.w3.org/TR/rdf12-turtle/#sec-mediaReg) is considered.

and also 

> In any case, the version announcement is merely a hint;

IIRC the argument was "that is the way it's done on the web" (sorry can't remember the details).

For protocol, there is a bit of covariant/contravariant here because it affects both the request and the response.

I would have though that the `VERSION` in the query is more likely to be correct for the query but for the response, it is the ability of the response handler may be less than the query which was given as a passed into a library.

kasei: > I would have though that the VERSION in the query is more likely to be correct for the query but for the response, it is the ability of the response handler may be less than the query which was given as a passed into a library.

Maybe `version` has different needs than dataset definition, but I think it's going to be a really strange place to be if we say protocol takes precedence in one case, and query syntax takes precedence in the other.

afs: Protocol has two versions:

`Content-type:` and `Accept`.

For input, `Content-type:`, we need to recall (from the WG) why Turtle says what it says.

FWIW I am not advocating one or the other way round. If pushed, I'd say "if they are different, issue a warning" and not define the precedence because it seems like that is an application "bug".

kasei: > Yes, but the same is true for protocol and GSP ... where the Turtle text applies.

I'm worried I'm not fully following the point here. I was specifically talking about SPARQL Query vs. Protocol. I think there are different considerations for the RDF side of things like Turtle where the media type is about the **response** body, while the Query/Protocol stuff is about the **request** body (and which to me seems similar enough to the dataset definition issue to at least consider what the impact of adding asymmetry would be).

kasei: GSP is an interesting in-between, because it's query- and update-adjacent, but the media types are going to be about the data/response body.

afs: If a query is POSTed, it's `Content-Type:` - which is also the update case.

If a query is sent with GET, there is no media type and no conflicit.

Then there is "Accept" which is about the response.

kasei: Yes, agreed. But the "Accept"/response case is mostly about RDF (there's also the SPARQL Result formats to think about; I'm not sure any of them have any relevant language yet about how to interpret conflicting versions). Since this PR comment was on the spec text about the SPARQL Query version announcement, I was mostly thinking along the lines of the Query/Update request body, and how to interpret that in the presence of a conflicting media-type-provided version via the Protocol.

TallTed: @kasei @afs —

I didn't think this was such a challenge, or I'd have immediately opened a new issue for it. Perhaps that should happen now.

I did just read somewhere else (sorry, 100+ issues & PRs in the past few hours) where a conflict between HTTP headers and body RDF data/SPARQL query was specified to return an error, rather than the protocol forcing one of these to always have precedence. It seems likely that should be echoed here.

PR #406, opened 2026-07-15
SPARQL vocab: rename "encode" into "encodeForUri"

Tpt:
The function is named `ENCODE_FOR_URI` in the SPARQL syntax and encode URIs

### DIFF
--- a/sparql-ns.ttl
+++ b/sparql-ns.ttl
@@ -317,8 +317,8 @@ sparql:replace rdf:type sparql:Function ;
     rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/#func-replace> ;
     .
 
-sparql:encode rdf:type sparql:Function ;
-    rdfs:comment "This function encodes a string using a specified method (e.g., URI-encoding), returning the encoded version." ;
+sparql:encodeForUri rdf:type sparql:Function ;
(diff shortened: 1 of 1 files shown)


rubensworks (approved):

afs (approved):

sparql-ns.ttl
```diff
@@ -317,8 +317,8 @@ sparql:replace rdf:type sparql:Function ;
     rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/#func-replace> ;
     .
 
-sparql:encode rdf:type sparql:Function ;
-    rdfs:comment "This function encodes a string using a specified method (e.g., URI-encoding), returning the encoded version." ;
+sparql:encodeForUri rdf:type sparql:Function ;
+    rdfs:comment "This function encodes a string using a specified method (here, URI-encoding), returning the encoded version." ;
```
TallTed, line 321: Is there a way to specify a different method, with the new function name? If not, the comment should be rephrased.

I also wonder, are there now (or will there soon be) other `encode___()` variants for other methods?

If the answer to either of my questions means more changes to the body, please change this comment to an issue, to track it to resolution.

Tpt: > Is there a way to specify a different method, with the new function name? 

No, not currently. 

I have opened #409 to rephrase

PR #410, opened 2026-07-25
Algebra: add an explicit "distinct" flag to Aggregation()

Tpt:
There was no parameter in the `Aggregation()` algebra construct for the `DISTINCT` flag

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/410.html" title="Last updated on Aug 6, 2026, 4:21 PM UTC (39aa8a0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/410/178f1c4...39aa8a0.html" title="Last updated on Aug 6, 2026, 4:21 PM UTC (39aa8a0)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -8827,10 +8827,11 @@ WHERE {
             |exprlist| is a nonempty sequence of <a href="#expressions">expressions</a> and
             |A| is an algebraic query expression.</li>
           <li id="defn_absAggregation">
-            <a href="#defn_absAggregation" class="absOp">Aggregation</a>(|exprlist|, |func|, |scalarvals|, |grp|)
+            <a href="#defn_absAggregation" class="absOp">Aggregation</a>(|exprlist|, |func|, |distinct|, |scalarvals|, |grp|)
             is an algebraic query expression if
             |exprlist| is a nonempty sequence of <a href="#expressions">expressions</a> or the asterisk character (*),
             |func| is a <a href="#setFunctions">set function</a>,
+            |distinct| is a boolean `true` or `false`,
             |scalarvals| is a partial function (which may be the empty function, i.e., with an empty domain), and
             |grp| is an algebraic query expression of the form <a href="#defn_absGroup" class="absOp">Group</a>(<var>exprlist'</var>, |A|) where
             <var>exprlist'</var> is a sequence of <a href="#expressions">expressions</a> and
@@ -8846,7 +8847,7 @@ WHERE {
             <var>A<sub>1</sub></var>, ..., <var>A<sub>|n|</sub></var> is
             a non-empty sequence of algebraic query expressions (i.e., |n| &geq; 1)
             such that every expression <var>A<sub>|i|</sub></var> in this sequence
-            is of the form <a href="#defn_absAggregation" class="absOp">Aggregation</a>(<var>exprlist<sub>|i|</sub></var>, <var>func<sub>|i|</sub></var>, <var>scalarvals<sub>|i|</sub></var>, <var>grp<sub>|i|</sub></var>)
+            is of the form <a href="#defn_absAggregation" class="absOp">Aggregation</a>(<var>exprlist<sub>|i|</sub></var>, <var>func<sub>|i|</sub></var>, <var>distinct<sub>|i|</sub></var>, <var>scalarvals<sub>|i|</sub></var>, <var>grp<sub>|i|</sub></var>)
             that is captured by <a href="#defn_absAggregation">the previous point</a>.</li>
           <li id="defn_absOrderBy">
             <a href="#defn_absOrderBy" class="absOp">OrderBy</a>(|A|, |condition|)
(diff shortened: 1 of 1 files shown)


Tpt:
The approach taken by this MR is to add an extra *distinct* parameter to *Aggregation()* that is either `true` or `false`. Not sure if it's the nicer way to archive the goal. Glad to change it to something else

spec/index.html
```diff
@@ -10523,20 +10529,21 @@ <h4>Aggregate Algebra</h4>
...
               <var>key<sub><var>m</var></sub></var>→<var>Ψ<sub><var>m</var></sub></var> } )<br>
-              &nbsp;&nbsp;&nbsp;= { (<var>key</var>, <var>F</var>(<var>Ψ</var>)) | <var>key</var> → <var>Ψ</var> in { <var>key<sub>1</sub></var>→<var>Ψ<sub>1</sub></var>, ...,
+              &nbsp;&nbsp;&nbsp;= { (<var>key</var>, <var>F</var>(<var>Ψ</var>, <var>distinct</var>)) | <var>key</var> → <var>Ψ</var> in { <var>key<sub>1</sub></var>→<var>Ψ<sub>1</sub></var>, ...,
               <var>key<sub><var>m</var></sub></var>→<var>Ψ<sub><var>m</var></sub></var> } }</p>
             <p>where<br>
               &nbsp;&nbsp;M(<var>Ψ</var>) = [ <a href="#defn_ListEval">ListEval</a>(<var>exprlist</var>, <var>μ</var>) | <var>μ</var> in <var>Ψ</var> ]<br>
-              &nbsp;&nbsp;F(<var>Ψ</var>) = <var>func</var>(M(<var>Ψ</var>), <var>scalarvals</var>), for non-<code>DISTINCT</code><br>
-              &nbsp;&nbsp;F(<var>Ψ</var>) = <var>func</var>(Dedup(M(<var>Ψ</var>)), <var>scalarvals</var>), for <code>DISTINCT</code></p>
+              &nbsp;&nbsp;F(<var>Ψ</var>, `false`) = <var>func</var>(M(<var>Ψ</var>), <var>scalarvals</var>) — non-<code>DISTINCT</code> case<br>
```
Tpt: I don't think the `—` to separate the definition from a comment is great, glad to change it to something better

spec/index.html
```diff
@@ -8827,10 +8827,11 @@ <h3>Algebraic Syntax</h3>
...
           <li id="defn_absAggregation">
-            <a href="#defn_absAggregation" class="absOp">Aggregation</a>(|exprlist|, |func|, |scalarvals|, |grp|)
+            <a href="#defn_absAggregation" class="absOp">Aggregation</a>(|exprlist|, |func|, |distinct|, |scalarvals|, |grp|)
             is an algebraic query expression if
             |exprlist| is a nonempty sequence of <a href="#expressions">expressions</a> or the asterisk character (*),
             |func| is a <a href="#setFunctions">set function</a>,
+            |distinct| is a boolean `true` or `false`,
             |scalarvals| is a partial function (which may be the empty function, i.e., with an empty domain), and
             |grp| is an algebraic query expression of the form <a href="#defn_absGroup" class="absOp">Group</a>(<var>exprlist'</var>, |A|) where
             <var>exprlist'</var> is a sequence of <a href="#expressions">expressions</a> and
```
TallTed, line 8837: The linebreaks that are seen here in the ReSpec source seem likely to help readers of the HTML output at least if not more than they help the authors (i.e., us). I suggest changing these into (possibly nested) list items, appropriately ordered and/or unordered. If that happens, most of the item ending punctuation (`,`, `and`, `, and`, etc.) should be removed or changed to just `.` ending each list item.

spec/index.html
TallTed: Any suggestion of a better separator? Maybe a newline plus a bit more indent plus whatever character you prefer to the em-dash?

I'm also noticing that some lines are indented with `&nbsp;&nbsp;` and others with `&nbsp;&nbsp;&nbsp;`. These are VERY small differences. If they should be different, I think the different should be larger — at *least* an additional `&nbsp;`, probably two or three more of these....

spec/index.html
```diff
@@ -10607,7 +10614,7 @@ <h4>Aggregate Algebra</h4>
             ?x.</p>
           <p>We produce <var>G</var> = <a href="#defn_algGroup" class="algFct">Group</a>((?x), <var>Ψ</var>) = { (1) → [<var>μ<sub>1</sub></var>, <var>μ<sub>2</sub></var>], (2) →
           [<var>μ<sub>3</sub></var>] }</p>
-          <p>And so <a href="#defn_algAggregation" class="algFct">Aggregation</a>((?y, ?z), ex:agg, {}, <var>G</var>) =<br>
+          <p>And so <a href="#defn_algAggregation" class="algFct">Aggregation</a>((?y, ?z), ex:agg, `false`, {}, <var>G</var>) =<br>
```
```suggestion
          <p>So <a href="#defn_algAggregation" class="algFct">Aggregation</a>((?y, ?z), ex:agg, `false`, {}, <var>G</var>) =<br>
```

spec/index.html
Tpt: Good point. Do it consistently would mean changing not only this definition but the ones around it, burring the substantive diff. Would you be fine if I do it in a follow-up?

spec/index.html
Tpt: Thank you!

> Maybe a newline plus a bit more indent plus whatever character you prefer to the em-dash?

Just tried it and was not convinced. The current setup seems to me the best I found yet.

spec/index.html
TallTed: Sure. Tracking in #412

spec/index.html
afs: Drop the comment. This is a definition.

The argument is called _distinct_. I think that's enough.

afs (approved):
(see comment as well)

spec/index.html
Tpt: Done: 459676738a1291ac8b62b29a8cf8a73ce3212d10

TallTed:
@afs — In https://github.com/w3c/sparql-query/pull/410#pullrequestreview-4835216358 you said —

> (see comment as well)

— but I cannot figure out to which comment you referred. Could you add a link? (You can right-click on the timestamps in these threads and "Copy Link Address" which you can then paste directly into a comment on a thread, as I did above in this comment, or make the href of any text in the body, as [I've done here](https://github.com/w3c/sparql-query/pull/410#pullrequestreview-4835216358).)

Tpt:
@TallTed I guess this one: https://github.com/w3c/sparql-query/pull/410#discussion_r3696156737

spec/index.html
```diff
@@ -9711,14 +9712,19 @@ <h5>Grouping and Aggregation</h5>
       End
   For each aggregate <var>R</var>(<var>args</var> ; <var>scalarvals</var>) now in <var>X</var>
```
hartig:
```suggestion
  For each aggregate <var>R</var>(<var>args</var> ; <var>scalarvals</var>) or <var>R</var>(DISTINCT <var>args</var> ; <var>scalarvals</var>) now in <var>X</var>
```

spec/index.html
hartig:
```suggestion
<p><b>Special Case:</b> when <code>COUNT</code> is used with the expression
<code>*</code>, then F(<var>Ψ</var>, <var>distinct</var>) is the cardinality of the group solution sequence,
i.e., F(<var>Ψ</var>, `false`)&nbsp;=&nbsp;<a href="#defn_Card">Card</a>(<var>Ψ</var>),
and F(<var>Ψ</var>, `true`)&nbsp;=&nbsp;<a href="#defn_Card">Card</a>(<a href="#defn_algDistinct">Distinct</a>(<var>Ψ</var>)).</p>
```

hartig (changes requested):
Good catch! Here are two edit suggestions to make this more accurate.

spec/index.html
Tpt: Done. Thank you for spotting this!

hartig (approved):

rubensworks (approved):

kasei (approved):

PR #411, opened 2026-07-27
Vocabulary: Remove `-distinct` variant of aggregation functions

### LINKED ISSUE #407: Are "distinct" variants of aggregate functions in the SPARQL vocabulary useful?
I am playing a bit with the SPARQL vocabulary and ended up reading this statement:

>> * Aggregate functions appear in two forms - with and without DISTINCT (are the distinct variants helpful?)
>
> Yes, the `DISTINCT` variants are helpful.
> _Originally posted by @TallTed citing @afs in https://github.com/w3c/sparql-query/issues/204#issuecomment-2813749985_

I tend to think these variants are an anti-pattern because it break alignment between built-in and user-provided functions.

Indeed, `DISTINCT` can be applied on "custom" aggregate functions identified by IRIs. This entails that `SUM(?x)` and `SUM(DISTINCT ?x)` are somehow "corresponding" to `sparql:agg-sum(?x)` and `sparql:agg-sum-distinct(?x)` but `ex:my_agg(DISTINCT ?x)` has no obvious correspondence dropping the `DISTINCT` modifier. Hence, implementations cannot just collapse the DISTINCT operation inside of the aggregate functions.

Moreover, SPARQL defines the aggregates independently from the `DISTINCT` operation that is applied *before*, leading to not having any properly defined semantic for these "distinct" variants.

@TallTed out of curiosity, what is your rational to consider them helpful?

Tpt:
Rational: DISTINCT is an operation done before invoking an aggregation function and is not part of it

Close #407 (see this issue for details)

### DIFF
--- a/sparql-ns.ttl
+++ b/sparql-ns.ttl
@@ -457,18 +457,8 @@ sparql:agg-count rdf:type sparql:Aggregate ;
     rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/> ;
     .
 
-sparql:agg-count-distinct rdf:type sparql:Aggregate ;
-    rdfs:comment "Aggregate function COUNT with DISTINCT" ;
-    rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/> ;
-    .
-
 sparql:agg-sum rdf:type sparql:Aggregate ;
     rdfs:comment "Aggregate function SUM" ;
-    rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/> ;
-    .
-
-sparql:agg-sum-distinct rdf:type sparql:Aggregate ;
-    rdfs:comment "Aggregate function SUM with DISTINCT" ;
     rdfs:isDefinedBy <http://www.w3.org/TR/sparql12-query/> ;
     .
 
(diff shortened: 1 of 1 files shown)


kasei (approved):

TallTed:
I think that voices in favor of this change are as important as voices against, so if you're reading this and have either opinion, please speak up. Neutral voices may also be important, at least to know you've considered the issue.

afs:
IMO
It is safer to not have them for now. 
We haven' published the document yet.

We can add them later (incremental), but removing them later is disruptive.

afs (approved):

hartig (approved):
I agree with @Tpt's and @kasei's comments in the corresponding issue (#407); the `DISTINCT` keyword used within an aggregate is meant to result in a duplicate removal that, conceptually, is applied before the actual aggregation function. In other words, there is only one [SUM](https://w3c.github.io/sparql-query/spec/#aggSum) function and that can be applied to an arbitrary solution sequence or to a duplicate free one. As a consequence, the IRI `sparql:agg-sum-distinct` should not be in the vocabulary, only `sparql:agg-sum`.

niklasl (approved):
TLDR: LGTM!

From my POV (as a user of SPARQL, with some additional, but mainly theoretical, cases for describing computations using this vocabulary),  I agree (like @hartig) with the arguments of @tpt and @kasei (in my understanding, the distinct solution modifier is part of the function invocation, applied before the aggregation occurs).

I also agree with @afs that *if* needs for composed versions of each aggregation function where to arise, they can be added later. Such an addition should then perhaps be considered in conjunction with particular approaches of describing invocation. For example SHACL Node Expressions [has a way](https://www.w3.org/TR/shacl12-node-expr/#DistinctExpression) to do this. While I believe such use cases are in scope here (i.e., SHACL using this vocabulary), rather than describing a complete SPARQL query like SPIN did, the latter also used a form of invocation description (with [`sp:distinct true`](https://www.w3.org/submissions/2011/SUBM-spin-sparql-20110222/#sp-Select)) for this purpose (similar, I believe, to oxigraph's internal representation [as mentioned](https://github.com/w3c/sparql-query/issues/407#issuecomment-5077377140) by @Tpt).

rubensworks (approved):

Issue #403, opened 2026-06-26
Unclear behavior of  `STRLEN` when U+0000 is present


Tpt:
U+0000 (the NUL character) is not part of [XML 1.1 `Char` production](https://www.w3.org/TR/xml11/#NT-Char) used by some XPATH functions like [`fn:string-length`](https://www.w3.org/TR/xpath-functions-31/#func-string-length) that SPARQL relies on for `STRLEN` definition. However, SPARQL relies on [RDF 1.2 string definition](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-string) that allows U+0000, leaving the behavior of `STRLEN` undefined when the input contains U+0000.

afs:
As I read it, it is not legal for `xsd:string` because `"\u0000"` is a datatype error. `xsd:string` (XML1.1) does not have U+FFFD or U+FFFE either (other non-characters are legal).

So it does not get to the call of [fn:string-length](https://www.w3.org/TR/xpath-functions-31/#func-string-length).

Lexical forms ([RDF String](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-string)) are all UTF-8 encodeable codepoints (i.e. excludes surrogates). 

The SPARQL "[string literal](https://www.w3.org/TR/sparql12-query/#dfn-string-literal)" definition could be extended now we have RDF String. All string related functions might need some revision.

`"\u0000"@fr` and `"\u0000"@fr--ltr` are legal! There are [no lexical form restrictions](https://www.w3.org/TR/rdf12-concepts/#literal-value) on in the value space mapping.

`STR("\u0000"@fr)` is a problem.

PR #415, opened 2026-08-24
Expand definition of STRLEN to allow NULL characters in lang-tagged strings.

kasei:
Proposed fix for #403. This only makes changes to the `STRLEN` definition (ignoring other string functions).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/415.html" title="Last updated on Aug 28, 2026, 9:10 PM UTC (b01295b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/415/50868ca...b01295b.html" title="Last updated on Aug 28, 2026, 9:10 PM UTC (b01295b)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -7093,8 +7093,8 @@ WHERE {
           <p>
             String functions operate over <a data-cite="RDF12-CONCEPTS#dfn-rdf-string">RDF strings</a>
             corresponding to the lexical value of the argument(s).
-            For <a data-cite="RDF12-CONCEPTS#dfn-language-tagged-string">language-tagged string</a>
-            and <a data-cite="RDF12-CONCEPTS#dfn-dir-lang-string">directional language-tagged string</a>,
+            For <a data-cite="RDF12-CONCEPTS#dfn-language-tagged-string">language-tagged strings</a>
+            and <a data-cite="RDF12-CONCEPTS#dfn-dir-lang-string">directional language-tagged strings</a>,
             this may include data which would not be valid as an <code>xsd:string</code>
             (for example, a string containing the NULL character <span class="monoplain">U+0000</span>).
             In such cases, the semantics of the underlying XPath string function are extended to support
--- a/spec/index.html
+++ b/spec/index.html
@@ -7094,12 +7094,23 @@ WHERE {
           <section id="func-strlen">
             <h5>STRLEN</h5>
             <pre class="prototype nohighlight"><span class="return">xsd:integer</span>  <span class="operator">STRLEN</span>(<span class="type">string literal</span> str)</pre>
-            <p>The <code>strlen</code> function corresponds to the XPath
+            <p>The <code>strlen</code> function returns an <code>xsd:integer</code> equal
+              to the length in codepoints of the <a data-cite="RDF12-CONCEPTS#dfn-lexical-form">lexical form</a>
+              of the literal. This corresponds to the XPath <a data-cite="XPATH-FUNCTIONS-31#func-string-length">fn:string-length</a>
+              function, with the exception that <code>strlen</code> covers codepoints outside those supported by
               <a data-cite="XPATH-FUNCTIONS-31#func-string-length">fn:string-length</a>
-              function and returns an
-              <code>xsd:integer</code> equal to the length in characters of the 
-              <a data-cite="RDF12-CONCEPTS#dfn-lexical-form">lexical form</a> of the
-              literal.
+              (covering all codepoints instead of just those corresponding to <a data-cite="XML11#charsets">XML Characters</a>).
+            </p>
+            
+            <p class="note">
+              The expanded support for codepoints beyond what
+              <a data-cite="XPATH-FUNCTIONS-31#func-string-length">fn:string-length</a>
+              supports allows for correctly computing the length of strings which include, for example,
+              the NULL character (<span class="monoplain">U+0000</span>). While the NULL character is not
+              valid in the value space of <code>xsd:string</code>, it is valid in a
+              <a data-cite="RDF12-CONCEPTS#dfn-language-tagged-string">language-tagged string</a>
+              or <a data-cite="RDF12-CONCEPTS#dfn-dir-lang-string">directional language-tagged string</a>
+              such as: "\u0000"@fr.
             </p>
             <div class="result">
               <table>
@@ -7119,6 +7130,10 @@ WHERE {
                   <tr>
                     <td><code>strlen("chat"^^xsd:string)</code></td>
                     <td>4</td>
+                  </tr>
+                  <tr>
+                    <td><code>strlen("\u0000"@fr)</code></td>
+                    <td>1</td>
                   </tr>
                 </tbody>
               </table>
--- a/spec/index.html
+++ b/spec/index.html
@@ -7090,6 +7090,14 @@ WHERE {
               </tbody>
             </table>
           </div>
+          <p>
+            String functions with operands which are <a data-cite="RDF12-CONCEPTS#dfn-language-tagged-string">language-tagged string</a>
+            or <a data-cite="RDF12-CONCEPTS#dfn-dir-lang-string">directional language-tagged string</a> may operate over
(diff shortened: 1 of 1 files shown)


afs:
Could this go at the point where "string literal" is defined [17.4.3 Functions on Strings](https://www.w3.org/TR/sparql12-query/#func-string)? Then it covers other functions.

hartig (approved):
Looks fine to me.

kasei:
@afs – 

I've updated to have the main text describing the extension beyond the XPath semantics in the "Functions on Strings" section, and included examples in most of the string functions to show use of strings with a NULL character.

A few things to call out that may need discussion:

* `STRENDS` - I've fixed what appears to be a bug in the examples: `strEnds("foobar"@en, "bar"@en) -> error`. I changed the args to be (xsd:string, lang-tagged literal) to match the similar example in `STRSTARTS`, and which should correctly be the error case.

* `REGEX` - I'm not sure if it's possible to express a NULL character in a regex pattern as defined by XML Schema. The definition of the pattern syntax seems contradictory in places. There's text that talks about  "XML characters" (which I think would preclude NULL), but then NULL is specifically included in what would match the `\w` escape:

  > `[#x0000-#x10FFFF]-[\p{P}\p{Z}\p{C}]` (all characters except the set of "punctuation", "separator" and "other" characters)

* `REPLACE` - the `pattern` and `replacement` arguments are currently specified as `xsd:string` which means you can't have a NULL directly in either one of them. Depending on the `REGEX` issue above, it may be possible to express a NULL in the pattern somehow, but it seems like we can't currently specify NULL in the replacement, even if the result is a lang-tagged literal.

spec/index.html
```diff
@@ -7090,17 +7090,24 @@ <h4>Functions on Strings</h4>
               </tbody>
             </table>
           </div>
+          <p>
+            String functions with operands which are <a data-cite="RDF12-CONCEPTS#dfn-language-tagged-string">language-tagged string</a>
+            or <a data-cite="RDF12-CONCEPTS#dfn-dir-lang-string">directional language-tagged string</a> may operate over
+            data which would not be valid as an <code>xsd:string</code> (for example, a language-tagged string containing
+            the NULL character <span class="monoplain">U+0000</span>). In such cases, the semantics of the underlying XPath
+            string function are extended to support the larger value space covering all codepoints instead of just those
```
Tpt: If I am not mistaken, surrogates are unicode codepoints. I am not sure we want to include them here. What about just stating that functions operates on [RDF strings](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-string) and not only `xsd:string`?

Tpt (commented):
Thank you for the MR.

> STRENDS - I've fixed what appears to be a bug in the example

Thanks!

> REGEX - I'm not sure if it's possible to express a NULL character in a regex pattern as defined by XML Schema. The definition of the pattern syntax seems contradictory in places. There's text that talks about "XML characters" (which I think would preclude NULL), but then NULL is specifically included in what would match the \w escape:

Imho I would allow NUL there for consistency but if you prefer to exclude it, I am also fine.

spec/index.html
kasei: I don't think surrogates can be expressed here. They are not allowed in UTF-8 and we disallow them in the code point escape sequences. Unless we want to take over the full definition of each of these functions, I think the links out to the XPath functions still makes sense, but then we need to explain how we handle more than just xsd:strings, which is what this text is trying to do.

kasei: Either way, I think it would be an issue for RDF-CONCEPTS if we wanted to make explicit assertions about what the value space of lang-tagged literals was regarding specific code points.

kasei:
> > REGEX - I'm not sure if it's possible to express a NULL character in a regex pattern as defined by XML Schema. The definition of the pattern syntax seems contradictory in places. There's text that talks about "XML characters" (which I think would preclude NULL), but then NULL is specifically included in what would match the \w escape:
> 
> Imho I would allow NUL there for consistency but if you prefer to exclude it, I am also fine.

I'm not sure what you mean by "allow NUL there". The reason we can't put NULL in a REGEX pattern directly is that the pattern argument is currently specified as an xsd:string. I don't think we can do anything about xsd:string not allowing NULLs. To have NULLs, the pattern would have to be a lang-tagged literal (with or without a direction), and then we'd have a weird situation of having to explain what to do with the language tag of the pattern (presumably just ignore it?).

spec/index.html
Tpt: Agreed. My comment was badly worded, the proper question should have been <q>What about just stating that functions operates on [RDF strings](https://www.w3.org/TR/rdf12-concepts/#dfn-rdf-string) and not only XML Char sequences?</q>.

Aside (please ignore):
It seems also we are touching a deeper issue. My understanding of RDF 1.2 concepts is that we are practically extending `xsd:string` definition to be any RDF string and not only sequences of XML `Char` (it would be quite weird that literals without datatype/language tag do not support all valid literal lexical forms).

kasei: Ah, I think I see. How about this (with links for definitions where appropriate)?

> String functions operate over RDF strings corresponding to the lexical value of the argument(s).
For language-tagged string and directional language-tagged string, this may include
data which would not be valid as an xsd:string (for example, a string containing
the NULL character U+0000). In such cases, the semantics of the underlying XPath
string function are extended to support the larger value space covering all Unicode scalar values
instead of just those corresponding to XML Characters.

Tpt: Sounds great to me! Thank you!

afs: In SPARQL, we should update [string literal](https://www.w3.org/TR/sparql12-query/#dfn-string-literal) (first bullet says "xsd:string" seems to be the trouble point) and may be note this at `STR()`.

We _could_ go further define `rdf:string` (in concepts?).

scalar values include U+FFFF and U+FFFE. It is a mystery to me why xsd:string disallows U+FFFF and U+FFFE. but allows U+7FFFE and U+7FFFF.

kasei: > In SPARQL, we should update string literal (first bullet says "xsd:string" seems to be the trouble point) and may be note this at STR().

What are you suggesting we update? I think the three bullet points still cover the full range of string literals(?). I'm not sure what `rdf:string` would be used for.

> scalar values include U+FFFF and U+FFFE. It is a mystery to me why xsd:string disallows U+FFFF and U+FFFE. but allows U+7FFFE and U+7FFFF.

Ignoring the vagaries of xsd:string, I think we do want these string functions to work over strings with any scalar value, right?

rubensworks (approved):

spec/index.html
```diff
@@ -7090,17 +7090,27 @@ <h4>Functions on Strings</h4>
               </tbody>
             </table>
           </div>
+          <p>
+            String functions operate over <a data-cite="RDF12-CONCEPTS#dfn-rdf-string">RDF strings</a>
+            corresponding to the lexical value of the argument(s).
+            For <a data-cite="RDF12-CONCEPTS#dfn-language-tagged-string">language-tagged string</a>
+            and <a data-cite="RDF12-CONCEPTS#dfn-dir-lang-string">directional language-tagged string</a>,
```
```suggestion
            For <a data-cite="RDF12-CONCEPTS#dfn-language-tagged-string">language-tagged strings</a>
            and <a data-cite="RDF12-CONCEPTS#dfn-dir-lang-string">directional language-tagged strings</a>,
```

spec/index.html
```diff
@@ -7090,17 +7090,27 @@ <h4>Functions on Strings</h4>
...
+            this may include data which would not be valid as an <code>xsd:string</code>
+            (for example, a string containing the NULL character <span class="monoplain">U+0000</span>).
+            In such cases, the semantics of the underlying XPath string function are extended to support
+            the larger value space covering all <a data-cite="I18N-GLOSSARY#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>
```
```suggestion
            the larger value space covering all <a data-cite="I18N-GLOSSARY#dfn-scalar-value" class="lint-ignore">Unicode scalar values</a>,
```

afs (approved):

Issue #402, opened 2026-06-26
Blank nodes in triple patterns within property path patterns


hartig:
While working on the evaluation semantics of property path patterns that contain triple patterns in their subject or object position I have come across a tricky issue: Consider the following query in which we have a property path pattern with a triple pattern in the object position, where that triple pattern contains a blank node that appears also in another, separate triple pattern.
```sparql
SELECT ?x ?y WHERE {
   :s  :p* <<( _:b :q ?x )>> .
   _:b :r ?y .
}
```
What would you expect to be the result of evaluating the given query over the following RDF graph?
```turtle
:s  :p :o .
:o  :p <<( :s1 :q "a" )>> .
:s2 :r "b" .
```
Do you expect the result to be i) the empty set or ii) a singleton set that contains the solution mapping _&mu;_ = {?x → "a", ?y → "b"} ?

The definition that I currently have for this case (which I explain a bit more below) produces the singleton set with _&mu;_, but I would find it much more intuitive if the result was the empty set.

Why would the empty set be more intuitive? Consider the following query, in which the only difference to the previous query is the property path expression.
```sparql
SELECT ?x ?y WHERE {
   :s :p/:p <<( _:b :q ?x )>> .
   _:b :r ?y .
}
```
For this query, the translation step in [Section 18.3.2.4 Translate Property Path Patterns](https://www.w3.org/TR/sparql12-query/#sparqlTranslatePathPatterns)) replaces the property path pattern by two triple patterns, resulting in the following query.
```sparql
SELECT ?x ?y WHERE {
   :s :p ?v .
   ?v :p <<( _:b :q ?x )>> .
   _:b :r ?y .
}
```
This rewritten query contains no more property path patterns but just a basic graph pattern consisting of three triple patterns (where one of them has another triple pattern in its object position). By the definition of [basic graph pattern matching](https://www.w3.org/TR/sparql12-query/#BGPsparql), the result of evaluating this rewritten query over the example graph above is the empty set. For this reason, it would feel totally non-intuitive to me if the result of the first query above was something else than the empty set.

But this brings us to the actual problem: How to define the [evaluation of ZeroOrMorePath](https://www.w3.org/TR/sparql12-query/#defn_evalZeroOrMorePath) (which is what the property path pattern of the first query uses) such that we would get the empty result for the first query?

[...]

PR #417, opened 2026-08-29
Removes blank nodes in the algebraic syntax

hartig:
As discussed in the context of #402 and [agreed during last week's SPARQL TF meeting](https://github.com/w3c/sparql-query/issues/402#issuecomment-5371961930), this PR removes blank nodes as possible elements of triple patterns, basic graph patterns, and property path patterns _within the algebraic syntax_.

While my initial idea was to implement this change directly by restricting the definitions of [property path patterns](https://www.w3.org/TR/sparql12-query/#defn_PropertyPathPattern) and [triple patterns](https://www.w3.org/TR/sparql12-query/#defn_TriplePattern) (and, thus, indirectly also of [basic graph patterns](https://www.w3.org/TR/sparql12-query/#sparqlBasicGraphPatterns)), I decided against that because these notions are mentioned pervasively throughout the informal parts of the document where there is no distinction between the formal concepts of triple patterns, BGPs, and property path patterns and the syntactic representation in the surface syntax. Therefore, I kept these definitions as they were and, instead, restricted their use within the algebraic syntax; more concretely, this means that BGPs in the algebraic syntax must be blank node free.

The PR then adds a new, blank-node-replacing step into [the translation procedure](https://www.w3.org/TR/sparql12-query/#translation), together with an adaptation of the step that handles SELECT expressions (which now has to make sure that the variables introduced by the new blank-node-replacing step are not projected).

The next change then is in the definition of BGP matching/evaluation, which has become much simpler because it needs to focus only on blank-node-free BGPs (i.e., no need anymore for considering [RDF instance mappings](https://www.w3.org/TR/rdf11-semantics/#dfn-instance) and introducing [pattern instance mappings](https://www.w3.org/TR/sparql11-query/#defn_PatternInstanceMapping)).

Finally, based on the new focus on only blank-node-free BGPs for the evaluation, some parts of the discussion in [Section 18.4.2 Treatment of Blank Nodes](https://www.w3.org/TR/sparql12-query/#BGPsparqlBNodes) and in [Section 18.6.3 Extending SPARQL Basic Graph Matching](https://www.w3.org/TR/sparql12-query/#sparqlBGPExtend) have become obsolete and are removed by this PR. @pfps please take a special look at the changes that the PR does in these sections.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/417.html" title="Last updated on Sep 10, 2026, 5:01 PM UTC (ba9847a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/417/db17cbc...ba9847a.html" title="Last updated on Sep 10, 2026, 5:01 PM UTC (ba9847a)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -9131,7 +9131,7 @@ WHERE {
               patterns.
             </li>
             <li>
-              <a href="#replaceBlankNodes">Replace blank nodes</a> by fresh variables.
+              <a href="#replaceBlankNodes">Replace blank nodes</a> with fresh variables.
             </li>
             <li>
               <a href="#sparqlTranslatePathExpressions">Translate property path expressions</a>
@@ -9313,7 +9313,7 @@ For each form FILTER(expr) in the group graph pattern
               and that neither the subject end point nor the object end point is a blank node
               (as <a href="#replaceBlankNodes">blank nodes have been replaced</a> before).
               The result of this step may be blank-node-free triple patterns or an
-              <a href="#defn_AlgebraicQueryExpression">algebraic query expressions</a>
+              <a href="#defn_AlgebraicQueryExpression">algebraic query expression</a>
               of the form <a href="#defn_absPath" class="absOp">Path</a>(...).</p>
             <p>Notes:</p>
             <ul>
@@ -9991,11 +9991,11 @@ The set <var>PV</var> is used later for projection (see Section&nbsp;<a href="#s
             because, during the <a href="#translation">translation to the algebraic syntax</a>,
             blank nodes have been replaced by fresh variables (see Section&nbsp;<a href="#replaceBlankNodes" class="sectionRef"></a>).
             Blank-node-free basic graph patterns can be instantiated
-            by replacing variables by RDF terms
+            by replacing variables with RDF terms
             according to a <a href="#defn_sparqlSolutionMapping">solution mapping</a>.</p>
           <p>For every basic graph pattern |B| and every solution mapping <var>μ</var>,
             <var>μ</var>(|B|) denotes the result of replacing every occurrence of variable |v|
-            in |B| by <var>μ</var>(|v|) for every variable |v| in dom(<var>μ</var>).</p>
+            in |B| with <var>μ</var>(|v|) for every variable |v| in dom(<var>μ</var>).</p>
           <div class="defn">
             <b>Definition: Evaluation of a Basic Graph Pattern</b>
             <p>Let |BGP| be a blank-node-free basic graph pattern and let |G| be an RDF graph.</p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -8603,6 +8603,8 @@ WHERE {
             <p>Triple patterns do not permit cycles
               (i.e., a triple pattern cannot be contained within itself).</p>
           </div>
+          <p>A triple pattern (|s|, |p|, |o|) is said to be <em>blank node free</em>
+            if neither |s| nor |o| is a blank node.</p>
 
           <div class="note">
             <p>This definition of Triple Pattern includes literal subjects. 
@@ -8636,6 +8638,8 @@ WHERE {
               <a href="#defn_TriplePattern">Triple Patterns</a>.</p>
           </div>
           <p>The empty graph pattern is a basic graph pattern which is the empty set.</p>
+          <p>A basic graph pattern is said to be <em>blank node free</em>
+            if all of its triple patterns are blank node free.</p>
         </section>
         <section id="sparqlPropertyPaths">
           <h4>Property Path Patterns</h4>
@@ -8761,7 +8765,7 @@ WHERE {
           is defined recursively as follows:</p>
         <ul>
           <li>
-            A <a href="#defn_BasicGraphPattern">basic graph pattern</a>
+            A blank-node-free <a href="#defn_BasicGraphPattern">basic graph pattern</a>
             is an algebraic query expression.</li>
           <li>
             A multiset of <a href="#defn_sparqlSolutionMapping">solution mappings</a>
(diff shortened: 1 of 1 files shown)


TallTed:
English pickiness -- "replace" takes "with" in active sentences, and "by" in passive sentences.

#### Rules of thumb:

- **_Active sentence:_** Use **replace + [old item] + with + [new item]**.

- **_Passive sentence:_** Use **[old item] + is/was replaced + by + [new item or agent]**.

I've not analyzed all of our documents to see how many "replace { with | by }" instances existed before this PR. It may be better to put this in a new WG issue and follow-on PRs for each document. Please feel free to do so.

hartig:
> English pickiness -- "replace" takes "with" in active sentences, and "by" in passive sentences. [...]

Thanks for pointing out this issue. I have applied this rule now to the text changed/add by this PR (commit  https://github.com/w3c/sparql-query/pull/417/commits/7917834a7f1e5ced209e1737ab1729409703dc58).

spec/index.html
```diff
@@ -8761,7 +8765,7 @@ <h3>Algebraic Syntax</h3>
           is defined recursively as follows:</p>
         <ul>
           <li>
-            A <a href="#defn_BasicGraphPattern">basic graph pattern</a>
+            A blank-node-free <a href="#defn_BasicGraphPattern">basic graph pattern</a>
```
afs: Above, there are no `-` in "blank-node-free".

kasei, line 8768: I find it a bit stylistically confusing that the definitions use the hyphen-free "blank node free", but the usage here is "blank-node-free".

spec/index.html
```diff
@@ -9173,6 +9186,36 @@ <h5>Expand Syntax Forms</h5>
...
+              </div>
+              <div class="algExample2">
+                ?s :p ?x .<br/>
+                ?s :q ?y .<br/>
+                ?y :q ?o .<br/>
+                ?o :p/:p/:p ?z .
+              </div>
+            </div>
+            <p>All the fresh variables introduced in this translation step are collected in a set
+              that is used later when processing SELECT expressions (see Section&nbsp;<a href="#sparqlSelectExpressions" class="sectionRef"></a>).</p>
```
kasei, line 9217: Can we name this set? As a point of comparison, the following section ("Collect FILTER Elements") explicitly declares the set `FS` which "is used later".

spec/index.html
```diff
@@ -9967,28 +10024,18 @@ <h4>Treatment of Blank Nodes</h4>
...
-            <a data-cite="RDF12-SEMANTICS#dfn-lean">lean</a>, and it allows logically equivalent datasets to
-            yield different query results.</p>
         </section>
       </section>
       <section id="PropertyPathPatterns">
         <h3>Property Path Patterns</h3>
         <p>This section defines the evaluation of <a href="#defn_PropertyPathPattern">property path
-            patterns</a>. A property path pattern consists of a subject endpoint (an RDF term or a variable), a
+            patterns</a>. A property path pattern consists of a subject endpoint (an IRI, a literal, a variable, or a triple pattern), a
           <a href="#defn_PropertyPathExpr">property path expression</a>, and an object endpoint.</p>
```
kasei, line 10039: Does "object endpoint" need the explicit enumeration of possible types like "subject endpoint" has?

kasei (approved):
Approved, but with a few minor comments/questions that might lead to some tweaks at your discretion.

Tpt (approved):
I have also implemented the written algorithm in Oxigraph and it seems to work properly

spec/index.html
hartig: > I find it a bit stylistically confusing that the definitions use the hyphen-free "blank node free", but the usage here is "blank-node-free".

I was thinking about this issue when I wrote these parts. My understanding was that, when used before the corresponding noun, the version with the hyphens is more correct / more common in English. (@TallTed what is your advise?)

Having said  that, in this particular part of the document we may avoid the issue altogether by changing

> * A blank-node-free basic graph pattern is an algebraic query expression.

to

> * A basic graph pattern that is blank node free is an algebraic query expression.

Yet, in other places where I wrote "blank-node-free basic graph pattern", it may not be possible to use this trick.

spec/index.html
hartig: Done (commit https://github.com/w3c/sparql-query/pull/417/commits/58ec37bdda2ffd0e463e491cc479f64cb99878d4)

spec/index.html
hartig: I was thinking the same but did not want to add new things with this PR, only change the existing things that needed to be changed. Given your comment, I have now added it nonetheless (commit https://github.com/w3c/sparql-query/pull/417/commits/8927357c60c97c232ddf72db0fe0fd5586c2f70b).

spec/index.html
kasei: Yes, as I continued reading, it became more apparent that there was some logic behind the different variants. I don't feel strongly about this, so change or leave it at your discretion.

afs (approved):

spec/index.html
hartig: > Above, there are no `-` in "blank-node-free".

We already have a discussion of this topic: https://github.com/w3c/sparql-query/pull/417#discussion_r3936470599

TallTed: I think consistent construction will help comprehension, and I think the hyphen-full "blank-node-free" is better than the hyphen-free "blank node free". 

"free of blank nodes" would be even better.

TallTed (changes requested):

rubensworks (approved):

pfps (approved):
looks good.

pfps:
Do you think that entailment regimes will need changes?

[Answering my own question]  Technically no, as the BGP blank node stuff will not be relevant, but the text should be updated, so actually yes.

hartig:
Regarding the question of "blank-node-free" and "blank node free" that came up in a couple of comments, I have done another pass of the changes of this PR and rephrased all mentions of both of these two phrases by "free of blank nodes" ([as suggested](https://github.com/w3c/sparql-query/pull/417#discussion_r3981295878) by @TallTed) -- see commit https://github.com/w3c/sparql-query/pull/417/commits/ba9847a51a3db1ee881a34df51c49c2a112cd367

hartig:
@pfps thanks for checking!

> Do you think that entailment regimes will need changes?

I assume you mean the [SPARQL 1.2 Entailment Regimes spec](https://www.w3.org/TR/sparql12-entailment/). In this case, the answer is yes. Everything in that document that considers the case of triple patterns or BGPs that contain blank nodes needs to be revised. Just taking a quick look, I notice, for instance, that [Section 1.1.3 (Preliminary Definitions)](https://www.w3.org/TR/sparql12-entailment/#prelims) defines the notion of a triple pattern with blank nodes.

Issue #392, opened 2026-06-22
interpretation of the ExprTripleTerm is ambiguous


lisp:
the common understanding of a BIND form would suggest that any variables must be already bound.

this constraint loses immediacy for an ExprTripleTerm, which is permitted to resemble a pattern to be matched against the dataset.
for example, in https://openhumanitiesdata.metajnl.com/articles/10.5334/johd.235 this conception applies to the later examples.

it may be that this carries over the behaviour attached to an EXISTS form.

which is intended to be the case?
is such a triple term to be reduced or scanned?

afs:
Expressions work on RDF terms and takes the variable values from the current row.

SPARQL follows RDF 1.2 - not RDF-star.
The referenced article uses RDF-star.

See https://www.w3.org/TR/sparql12-query/#func-triple

lisp:
the referenced passage does suggest an equivalence between the triple constructor and the triple term, but does not indicate that it is by definition.

> Expressions work on RDF terms and takes the variable values from the current row.

unless the operator is exists.  
in which case, otherwise unbound variables contribute as pattern elements.

> The referenced article uses RDF-star.

the issue is a matter of concept, not editorial.

afs:
> the referenced passage does suggest an equivalence between the triple constructor and the triple term, but does not indicate that it is by definition.

It is normative.

> unless the operator is exists.

which uses `{  }` to carry a pattern

kasei:
The spec already says:

> Apart from the functional forms BOUND, COALESCE, IF, IN, NOT IN, logical-or (||), logical-and (&&), logical-not (!), NOT EXISTS, EXISTS, and EBV, all functions operate on RDF Terms. Functions produce an error if any argument is unbound.

I think what this might miss is a top-level expression (without a function being involved) that either is or contains (in the case of an ExprTripleTerm) a variable that is unbound.

Tpt:
> I think what this might miss is a top-level expression (without a function being involved) that either is or contains (in the case of an ExprTripleTerm) a variable that is unbound.

Indeed, it seems that the evaluation of a variable is not defined anywhere in the section. I have opened #392

> or contains (in the case of an ExprTripleTerm)

Do we need this case? If I understand the spec correctly `ExprTripleTerm` is just a syntactic sugar for the `TRIPLE` function

TallTed:
[...]

PR #419, opened 2026-09-10
Expressions: define variable evaluation

Tpt:
I don't think it was written anywhere

See #392

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/pull/419.html" title="Last updated on Sep 12, 2026, 2:25 PM UTC (8f053f4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/sparql-query/419/17ed801...8f053f4.html" title="Last updated on Sep 12, 2026, 2:25 PM UTC (8f053f4)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -4809,13 +4809,6 @@ WHERE { ?annot  a:annotates  &lt;http://www.w3.org/TR/rdf-sparql-query/&gt; .
         </p>
         <ul>
           <li>
-            If the SPARQL expression is a variable
-            and the solution mapping with respect to which the expression is evaluated contains that variable,
-            then the result of this evaluation is the RDF term that the solution mapping assigns to that variable.
-            If the SPARQL expression is a variable that the solution mapping does not contain,
-            then the result of evaluating the expression is an error.
-          </li>
-          <li>
             Unlike functions in XPath/XQuery, functions in SPARQL do not
             process node sequences.
           </li>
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -4808,6 +4808,10 @@ <h3>Expression Evaluation</h3>
           models between XPath/XQuery and SPARQL:
         </p>
         <ul>
+          <li>
+            Variables are evaluated to the corresponding RDF term in the <a href="#defn_sparqlSolutionMapping">solution mapping</a>
+            or an error if the variable is not in it.
```
hartig, line 4813:
```suggestion
            If the SPARQL expression is a variable, the result of evaluating it
            with respect to a <a href="#defn_sparqlSolutionMapping">solution mapping</a>
            is the RDF term that the solution mapping assigns to that variable,
            or an error if the the solution mapping does not contain the variable.
```

hartig (changes requested):
In addition to my edit suggestion, I propose to move this new bullet point somewhere else within the list of bullet points, because it does not fit so nicely as the first bullet point after the sentence that advertises the list to be about rules related to the differences between XPath/XQuery and SPARQL. Maybe you can just make it the last bullet point in the list?

spec/index.html
TallTed: As edited above (and maintained in my suggestion below), this says that the variable is `set ... to an error if the solution mapping does not contain the variable`. Is this intended? It seems more likely that the variable should not be set to any value, and the error raised somewhere, though I'm not sure where.

I *think* my rephrasing is clearer (though it still has the confusing "set to an error" I mention above), but it may need some further work from me or others, if it's unclear to anyone reading this PR and this change suggestion.

Also note that you have a duplicate `the` in line 4815, which I've removed in the following.
```suggestion
            If the SPARQL expression is a variable, it is set to the RDF term
            which the solution mapping assigns to it as a result of evaluating
            it with respect to a
            <a href="#defn_sparqlSolutionMapping">solution mapping</a>,
            or an error if the solution mapping does not contain the variable.
```

hartig: It is not the case that "the variable is set to" anything.

Here is another attempt:

```suggestion
            If the SPARQL expression is a variable
            and the solution mapping with respect to which the expression is evaluated contains that variable,
            then the result of this evaluation is the RDF term that the solution mapping assigns to that variable.
            If the SPARQL expression is a variable that the solution mapping does not contain,
            then the result of evaluating the expression is an error.
```

Tpt:
> I propose to move this new bullet point somewhere else within the list of bullet points

@hartig done

hartig (approved):

kasei (approved):

rubensworks (approved):

PR #424, opened 2026-09-17
add comparison of lateral and deep injection exists approaches

lisp:
the exists-lateral-join-deep-injection-modes.md document compares an exists implementation which performs a lateral join when possible with the deep injection approach described in Defining_the_DEEP_INJECTION_approach_for_EXISTS.md.


TallTed:
Please confirm whether the `Defining_the_DEEP_INJECTION_approach_for_EXISTS.md` to which you refer is that found here: [`https://github.com/w3c/sparql-query/blob/main/discussion/Defining_the_DEEP_INJECTION_approach_for_EXISTS.md`](https://github.com/w3c/sparql-query/blob/main/discussion/Defining_the_DEEP_INJECTION_approach_for_EXISTS.md).

I believe but am not certain that the `exists-lateral-join-deep-injection-modes.md` you intend is [`https://github.com/w3c/sparql-query/blob/a52c806a0211c278141d3363f98cc0a320c4046c/discussion/exists-lateral-join-deep-injection-modes.md`](https://github.com/w3c/sparql-query/blob/a52c806a0211c278141d3363f98cc0a320c4046c/discussion/exists-lateral-join-deep-injection-modes.md).

It would be helpful if you could link directly to such documents. There is no guarantee we will find the same document(s) nor revision(s) thereof as you intended, if we must search them out ourselves.

hartig:
Merging this PR. No need to review it explicitly as it adds a separate document to the repo, rather than changing anything in the spec.

@lisp Thanks for taking the time to create this analysis!! I hope to find some time to look at this in detail...

